PR #62, opened 2021-03-10
Move Group C++ Interface

MarqRazz:
This is the initial port of the Move Group C++ Interface tutorial and addresses #2 . This tutorial is very similar to [moveit2_demo_nodes/run_move_group](https://github.com/ros-planning/moveit2/tree/main/moveit_demo_nodes/run_move_group) which is where I grabbed the launch files.

To test this code run the following:
`ros2 launch moveit2_tutorials move_group.launch.py`
`ros2 launch moveit2_tutorials move_group_interface_tutorial.launch.py`

Items that I still need to address:
- [x] Clean up code and get the markers working with rviz_visual_tools until moveit_visual_tools is ported
- [x] Generate the webpage and see how it looks
- [x] Add more documentation
- [x] Get the collision scene examples working better (they tend to fail to plan)
- [x] Add to the readme and webpage that you need `xterm` installed

### Checklist


doc/move_group_interface/CMakeLists.txt
```diff
@@ -1,8 +1,16 @@
...
-install(DIRECTORY launch DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
+install(TARGETS move_group_interface_tutorial
+  DESTINATION lib/${PROJECT_NAME}
+)
+install(DIRECTORY launch
+  DESTINATION share/${PROJECT_NAME}
+)
+install(DIRECTORY config
+  DESTINATION share/${PROJECT_NAME}
+)
```
JafarAbdi, line 16: We should use the YAML files from moveit_resources

```suggestion
```

doc/move_group_interface/launch/move_group.launch.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Publish TF
+    robot_state_publisher = Node(package='robot_state_publisher',
+                                 executable='robot_state_publisher',
+                                 name='robot_state_publisher',
+                                 output='both',
+                                 parameters=[robot_description])
+
+    # Fake joint driver
+    fake_joint_driver_node = Node(package='fake_joint_driver',
```
JafarAbdi, line 102: We replaced this node with the one from ros2_control [see](https://github.com/ros-planning/moveit2/blob/main/moveit_demo_nodes/run_move_group/launch/run_move_group.launch.py#L144)

JafarAbdi (changes requested):

MarqRazz:
Changes were added in #69

Issue #1, opened 2021-02-04
Port "MoveIt Quickstart in RViz" tutorial


bostoncleek:
Port tutorial to ROS2.

lorepieri8:
With #63 this should be closed right?
Also, the website warning need to be removed: http://moveit2_tutorials.picknik.ai/doc/quickstart_in_rviz/quickstart_in_rviz_tutorial.html

vatanaksoytezer:
#63 closes this.

PR #63, opened 2021-03-12
Port "Quickstart in RViz" Tutorial

sea-bass:
"Quickstart in RViz" Tutorial has been ported to work with ROS2. 

The key update to the code is the addition of a ROS2 compatible `demo.launch.py` and associated RViz configurations (empty and filled out) that also work with ROS2. Previously, this relied on the `moveit_resources` repository but I have moved these assets within the actual `doc/quickstart_in_rviz` folder, which has led to a few changes in the `CMakeLists.txt`.

```
ros2 launch moveit2_tutorials demo.launch.py rviz_tutorial:=true
```

The tutorial itself has been updated to correct topic names and sliders that did not originally work by exactly following the ROS1 instructions. Specifically,

* The `move_group/monitored_planning_scene` topic did not work, but `/monitored_planning_scene` did.
* The `move_group/display_planned_path` topic similarly did not work, but `/display_planned_path` did.
* Replace the **MotionPlanning - Slider** with **Trajectory - Trajectory Slider**.
* The **RVizVisualToolsGui** panel is not available from the **Panels** drop-down menu -- you actually have to select **Add New Panel** and look for it from the full list.

Closes Issue #1.


MarqRazz (approved):
Sorry it took so long to get this reviewed @sea-bass! Thanks for your contributions!

PR #65, opened 2021-03-18
Add conditional installation of sphinx to build_locally.sh

MarqRazz:
I could also try and check if sphinx is installed with
```test -x `which sphinx-build` || pip install sphinx==1.8.5```

I am specifically using pip so that I can ensure we are using the same version as ROS1, but if you would prefer we use apt we just need to double check which version gets installed.

### CODE REVIEW

File: build_locally.sh [hunk 1]
```diff
@@ -1,5 +1,12 @@
 #!/bin/sh
 
+# Install sphinx-build if it isn't there yet
+if ! type "sphinx-build" > /dev/null
```
MarqRazz, line 4: I tried to use the same logic as [moveit_tutorials](https://github.com/ros-planning/moveit_tutorials/blob/master/build_locally.sh#L4) but the script would always attempt to reinstall sphinx.

`if ! command -v sphinx-build &> /dev/null`

vatanaksoytezer (approved):
Tested in my local, gets the job done. We can merge

MarqRazz:
@JafarAbdi would you mind taking at look at this and we can get it merged?

PR #70, opened 2021-04-14
Enable building only main branch

JafarAbdi:
### Description

We need to enable `Build pushed branches` from Travis to deploy the website on new commits, this PR will restrict the deployment to only use the main branch

- [x] Enable `Build pushed branches` from [Travis settings](https://travis-ci.com/github/ros-planning/moveit2_tutorials/settings)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,10 @@ compiler: gcc
 notifications:
   email: true
 
+branches:
+  only:
+    - main
+
 env:
   global:
     - SCRIPT=.moveit_ci/travis.sh


vatanaksoytezer:
@JafarAbdi I think we need to add /cpp/ after html/ part of the broken links to pass travis. Should I open a PR with that?

JafarAbdi:
> @JafarAbdi I think we need to add /cpp/ after html/ part of the broken links to pass travis. Should I open a PR with that?

Yes, please do, thanks!

auman66 (approved):

PR #69, opened 2021-04-14
Fix move group namespace and demo.launch.py

vatanaksoytezer:
- Ports #62, #63 and #64 to the new ROS2 conventions.
- Fixes #68.
- Includes move_group namespace fixes (see https://github.com/ros-planning/moveit2/pull/420)
- Add pre-commit

PR #71, opened 2021-04-14
Fix broken moveit_core api html links

vatanaksoytezer:
Fixing the broken links to pass travis. Relates to fails from #69 and #70.

This was due to moveit_core api page being moved from /html/ to /html/cpp/ such as:

http://docs.ros.org/en/noetic/api/moveit_core/html/classmoveit_1_1core_1_1JointModelGroup.html moved to
http://docs.ros.org/en/noetic/api/moveit_core/html/cpp/classmoveit_1_1core_1_1JointModelGroup.html


JafarAbdi (approved):
Thanks!

vatanaksoytezer:
Seems to be passing @JafarAbdi, you can merge if it looks good.

PR #90, opened 2021-06-07
Cleanup CI

vatanaksoytezer:
Cleans up the GH action configuration files


moveit2_tutorials.repos
```diff
@@ -1,9 +1,5 @@
 repositories:
-  moveit2_tutorials:
```
MarqRazz, line 2: Do any of the instructions need to be updated by removing this? I think the readme expects the repo itself to be included in this file. I will try to dig around a little and see.

MarqRazz (commented):
What did we replace `htmlcheck.yaml` with?

moveit2_tutorials.repos
vatanaksoytezer: Good catch, fixing it.

README.md
```diff
@@ -42,6 +42,7 @@ Download Moveit2_tutorials Source Code
 
     wget https://raw.githubusercontent.com/ros-planning/moveit2_tutorials/main/moveit2_tutorials.repos
     vcs import < moveit2_tutorials.repos
+    git clone https://github.com/ros-planning/moveit2_tutorials.git
     rosdep install -r --from-paths . --ignore-src --rosdistro foxy -y
```
MarqRazz, line 46: Sorry I didnt say but we should change this to clone `moveit2_tutorials` and then extract the `.repos` from that folder and not use `wget`.

```
git clone https://github.com/ros-planning/moveit2_tutorials.git
vcs import < moveit2_tutorials/moveit2_tutorials.repos
rosdep install -r --from-paths . --ignore-src --rosdistro foxy -y
```

vatanaksoytezer: Sure

MarqRazz:
Vatan, any idea on why the `CheckBuildHTML` action is not reporting success?
image

vatanaksoytezer:
> Vatan, any idea on why the `CheckBuildHTML` action is not reporting success?
> image

I changed the name of it to Build Pages

vatanaksoytezer:
> What did we replace `htmlcheck.yaml` with?

Initially I thought of removing it with parallel to what we did in moveit_tutorials with Robert, but I think keeping it does not harm and reasonable.

MarqRazz:
> > Vatan, any idea on why the `CheckBuildHTML` action is not reporting success?
> > image
> 
> I changed the name of it to Build Pages

Do you know why `CheckBuildHTML` is still in the Action tests list and showing the yellow running state? Is it because that test ran when you started the PR so now it always expects it?

If you changed the name do we need to update the badges on the README?

vatanaksoytezer:
> Is it because that test ran when you started the PR so now it always expects it?

That is the reason. Once this PR is merged the action will be replaced with the new action and will no longer be expected. We would need to remove / delete the previous runs in the action list in Github for full cleanup though. I will manually delete that action from the list after the merge. The name of the file has not been changed, so no need for badge updates. The internal name of the action is the that got changed

MarqRazz (approved):
It looks like we dont have a badge for the HTML check if you feel like adding one.

Thanks for getting this done it looks much cleaner.

vatanaksoytezer:
It should be much cleaner with the last commit now. We need someone with admin access to change the required actions to new ones to merge this.

.github/workflows/industrial_ci_action.yml
```diff
@@ -11,85 +11,38 @@ on:
...
-          df -h
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: cache upstream_ws
         uses: pat-s/always-upload-cache@v2.1.3
         with:
           path: ${{ env.BASEDIR }}/upstream_ws
-          key: upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2.repos') }}-${{ github.run_id }}
+          key: upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2_tutorials.repos.repos') }}-${{ github.run_id }}
```
rhaschke, line 40: This seems to be the wrong file name: *.repos.repos
```suggestion
          key: upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2_tutorials.repos') }}-${{ github.run_id }}
```

.github/workflows/industrial_ci_action.yml
```diff
@@ -11,85 +11,38 @@ on:
...
           restore-keys: |
-            upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2.repos') }}
+            upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2_tutorials.repos.repos') }}
```
rhaschke, line 42: Same here.
```suggestion
            upstream_ws-${{ env.CACHE_PREFIX }}-${{ hashFiles('moveit2_tutorials.repos') }}
```

rhaschke (approved):
Some small issue(s) to fix remain. Otherwise, I approve.

rhaschke:
@henningkayser, I guess you have admin rights here to remove CheckBuildHTML as a required CI job.

MarqRazz:
So I removed `CheckBuildHTML` from the list of "Status checks that are required" under the branch protection rules but still cant get the option to merge it.
image

I can try to disable `Require status checks to pass before merging` and re-enable it after if you guys think that is a good option.

vatanaksoytezer:
> I can try to disable `Require status checks to pass before merging` and re-enable it after if you guys think that is a good option.

That looks reasonable we can enable it right back

PR #96, opened 2021-07-10
Draw attention to return value of MoveGroupInterface::setJointValueTarget

nbbrooks:
### Description

This updates the tutorial to make users aware the existence and role of the MoveGroupInterface::setJointValueTarget boolean return.

Move group currently will allow setJointValueTarget to set a joint target outside of joint limits and MoveGroupInterface::plan will return a success (with a warning of "Joint <> is constrained to be above the maximum bounds. Assuming maximum bounds instead."). This is unexpected behavior by some users and this update to the tutorial will make it more clear where they should check to catch this kind of behavior.

It also makes a small clarification to the xterm usage.

### Checklist

### DIFF
--- a/doc/examples/move_group_interface/src/move_group_interface_tutorial.cpp
+++ b/doc/examples/move_group_interface/src/move_group_interface_tutorial.cpp
@@ -188,7 +188,8 @@ int main(int argc, char** argv)
   // Now, let's modify one of the joints, plan to the new joint space goal, and visualize the plan.
   joint_group_positions[0] = -1.0;  // radians
   bool within_bounds = move_group.setJointValueTarget(joint_group_positions);
-  if(!within_bounds) {
+  if (!within_bounds)
+  {
     RCLCPP_WARN(LOGGER, "Target joint position(s) were outside of limits, but we will plan and clamp to the limits ");
   }
 
--- a/doc/examples/move_group_interface/src/move_group_interface_tutorial.cpp
+++ b/doc/examples/move_group_interface/src/move_group_interface_tutorial.cpp
@@ -187,7 +187,10 @@ int main(int argc, char** argv)
 
   // Now, let's modify one of the joints, plan to the new joint space goal, and visualize the plan.
   joint_group_positions[0] = -1.0;  // radians
-  move_group.setJointValueTarget(joint_group_positions);
+  bool within_bounds = move_group.setJointValueTarget(joint_group_positions);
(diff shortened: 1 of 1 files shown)


AndyZe:
IMO this is more of a bug that should be patched rather than trying to fix with documentation.

vatanaksoytezer:
> IMO this is more of a bug that should be patched rather than trying to fix with documentation.

@nbbrooks I agree Andy on this one I think we need a fix in moveit2 rather then fixing the documentation. The comment on moveit_visual_tools / rviz_visual_tools is correct and I think we need to change this in multiple tutorials. I will open a PR soon on that, but I feel like I can add that rviz functionality to ROS2. Were there any reason on why that hasn't ported to ROS2?

nbbrooks:
I would prefer this be fixed as well. I will backport this to MoveIt 1 tutorials instead.

Maybe @henningkayser knows more about the remote control port status.

doc/move_group_interface/src/move_group_interface_tutorial.cpp
```diff
@@ -201,7 +201,10 @@ int main(int argc, char** argv)
 
   // Now, let's modify one of the joints, plan to the new joint space goal and visualize the plan.
   joint_group_positions[0] = -1.0;  // radians
-  move_group.setJointValueTarget(joint_group_positions);
+  bool within_bounds = move_group.setJointValueTarget(joint_group_positions);
+  if(!within_bounds) {
+    RCLCPP_WARN(LOGGER, "Target joint position(s) were outside of limits, but we will plan and clamp to the limits ");
```
henningkayser, line 206: This warning (debug?) could be useful in MGI as well.

henningkayser (commented):
I think the addition to the tutorial makes sense. I wouldn't go so far to call it a bug right away, I mean it's documented and handled behavior, just not very intuitive. At least I would check for side effects before changing the behavior in `kinematic_constraints`.

davetcoleman:
@tylerjw did a big refactor of this tutorial, right? I bet this PR is no longer relevant

doc/move_group_interface/src/move_group_interface_tutorial.cpp
tylerjw: Maybe the return value from this function should just be marked [[nodiscard]]?

JafarAbdi (approved):

PR #103, opened 2021-08-11
Update installation instructions, and minor claenup

vatanaksoytezer:
Updates the installation instructions to point website and make the instructions distro-independent.

### DIFF
--- a/doc/getting_started/getting_started.rst
+++ b/doc/getting_started/getting_started.rst
@@ -1,24 +1,24 @@
 Getting Started
 ===============
 
-This tutorial will install MoveIt2 and create a workspace sandbox to run the tutorials and example robot.
+This tutorial will install MoveIt 2 and create a workspace sandbox to run the tutorials and example robot.
 
-Install ROS2 and Colcon
+Install ROS 2 and Colcon
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-`Install ROS2 Foxy <https://docs.ros.org/en/foxy/Installation.html>`_.
-It is easy to miss steps when going through the ROS2 installation tutorial. If you run into errors in the next few steps, a good place to start is to go back and make sure you have installed ROS2 correctly.
+`Install ROS 2 Foxy <https://docs.ros.org/en/foxy/Installation.html>`_.
+It is easy to miss steps when going through the ROS 2 installation tutorial. If you run into errors in the next few steps, a good place to start is to go back and make sure you have installed ROS 2 correctly.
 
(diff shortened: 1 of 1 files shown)


tylerjw (approved):

doc/getting_started/getting_started.rst
```diff
@@ -1,49 +1,51 @@
...
 
   sudo apt install python3-vcstool
 
-Create A Colcon Workspace and Download MoveIt2 Source
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-These tutorials rely on the master branch of MoveIt2, which requires a build from source.
-You will need to have a `colcon <https://docs.ros.org/en/foxy/Tutorials/Colcon-Tutorial.html#install-colcon>`_ workspace setup: ::
+Create A Colcon Workspace and Download MoveIt 2 Source
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+These tutorials rely on the master branch of MoveIt 2, which can be built from source or binaries. Please make sure you install MoveIt 2 for your distribution of choice by following the instructions from `the official MoveIt website <https://moveit.ros.org/install-moveit2/source/>`_.
```
tylerjw, line 31:
```suggestion
These tutorials rely on the ROS 2 project MoveIt 2, which can be built from source or installed from binaries. Please make sure you install or build MoveIt 2 for your distribution of choice by following the instructions from `the official MoveIt website <https://moveit.ros.org/install-moveit2/source/>`_.
```

We don't have a master branch of MoveIt 2.  It is a main branch.  These tutorials do not necessarily depend on that branch.  In the future we might need a split between ROS versions of MoveIt 2 for the tutorials but until then we should just say it depends on MoveIt 2.

Secondly, the instructions on the website do not mention you need to use the `foxy` branch of moveit2 if you are on foxy.  Lastly, the website does not offer instructions for binary installs.

I think we should encourage users on foxy and galactic to install the binaries and users who want to build from source to use rolling.  Thoughts?

tylerjw (commented):
Looked more closely at this and left a comment.

doc/getting_started/getting_started.rst
vatanaksoytezer: I don't think you would need to use foxy branch for tutorials it compiles just fine on Rolling as well 👀 . We can always encourage using binaries (except for the time being rolling binaries I think are broken from Duration bug, and galactic is not synced yet.) Source builds should be only for contributing or to get unreleased features (namely between syncs).

tylerjw (approved):

PR #113, opened 2021-09-14
[Servo] use the default launch file from the pkg

AndyZe:
[This PR](https://github.com/ros-planning/moveit2/pull/621) added a new example launch file. It is nice in that it shows how to launch Servo as a node component or as a standalone node.

So, now the tutorial can use that launch file and we can delete some of the launch files in this repo.

This launch file was not used anywhere AFAICT so I deleted it:  `servo_server_panda.launch.py`


facontidavide:
In your branch I can see the file **servo_cpp_interface_demo.launch.py** but your instructions are:

> ros2 launch moveit_servo servo_example.launch.py

facontidavide:
My concern with this is that it affects the MoveIt2 Foxy documentation and these particular changes were not ported to Foxy (main only).

Therefore:

- Either we backport the changes merged in **moveit_servo**  to Foxy
or
- we merge this PR to the version of ROS2 where `moveit_ros/moveit_servo/launch/servo_example.launch.py` can be found.

AndyZe:
Thanks for the review @facontidavide! I'll get the ball rolling on backporting that PR to Foxy.

AndyZe:
@vatanaksoytezer @JafarAbdi  it looks like we can merge this now since it will not affect Foxy. I would appreciate that to simplify launch file maintenance.

Issue #125, opened 2021-10-04
foxy branch


tylerjw:
### Description

Relating to this:
* https://github.com/ros-planning/moveit2/issues/716

I think it would be a much better first-time user experience if there was a foxy branch of the tutorials that had these changes from the main branch:

* Default for the landing page of moveit2_tutorials
* Based on the debian Foxy release of MoveIt2
* All non-ported and tutorials that depend on source builds are removed from the foxy branch
* Link to the main branch of tutorials that could then be based on Rolling

@JafarAbdi do you know what we would need to do to get a new branch here to generate into the website?

tylerjw:
* https://github.com/ros-planning/moveit2_tutorials/pull/126
*

PR #126, opened 2021-10-05
Deploy from foxy branch

tylerjw:
### Description

This is the first step in accomplishing this:

* https://github.com/ros-planning/moveit2_tutorials/issues/125

I already pushed this into the foxy branch, this just cleans up the main branch (so CI won't deploy the main branch any more).


vatanaksoytezer (changes requested):
We should add foxy branch to industrial CI as well: https://github.com/ros-planning/moveit2_tutorials/blob/main/.github/workflows/industrial_ci_action.yml

tylerjw:
> We should add foxy branch to industrial CI as well: https://github.com/ros-planning/moveit2_tutorials/blob/main/.github/workflows/industrial_ci_action.yml

I updated this to separate the CI runs for foxy and main.  I also updated the readme badges.

vatanaksoytezer (approved):

PR #127, opened 2021-10-05
Cleanup getting started for foxy

tylerjw:
### Description

Cleaning up the instructions for installing Foxy to make them simpler.

In support of:

* https://github.com/ros-planning/moveit2/issues/716

### DIFF
--- a/doc/getting_started/getting_started.rst
+++ b/doc/getting_started/getting_started.rst
@@ -6,7 +6,11 @@ This tutorial will install MoveIt 2 and create a workspace sandbox to run the tu
 Install ROS 2 and Colcon
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 `Install ROS 2 Foxy <https://docs.ros.org/en/foxy/Installation.html>`_.
-It is easy to miss steps when going through the ROS 2 installation tutorial. If you run into errors in the next few steps, a good place to start is to go back and make sure you have installed ROS 2 correctly.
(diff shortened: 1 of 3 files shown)


doc/getting_started/getting_started.rst
```diff
@@ -18,18 +18,20 @@ Once you have ROS 2 installed, make sure you have the most up to date packages:
   sudo apt update
   sudo apt dist-upgrade
 
-Install `Colcon <https://docs.ros.org/en/foxy/Tutorials/Colcon-Tutorial.html#install-colcon>`_ the ROS 2 build system: ::
+Install `Colcon <https://docs.ros.org/en/foxy/Tutorials/Colcon-Tutorial.html#install-colcon>`_ the ROS 2 build system with `mixin <https://github.com/colcon/colcon-mixin-repository>`_: ::
```
vatanaksoytezer, line 21: I use and like mixin, but I am not %100 sure we should recommend it as default. Maybe we can add a section like "hey there is a nice tool called mixin that most devs use."

doc/getting_started/getting_started.rst
```diff
@@ -38,19 +40,19 @@ Download MoveIt 2 Tutorials Source
 Move into your colcon workspace and pull the MoveIt 2 tutorials source: ::
 
   cd ~/ws_moveit2/src
-  git clone https://github.com/ros-planning/moveit2_tutorials.git
+  git clone https://github.com/ros-planning/moveit2_tutorials.git -b foxy
```
vatanaksoytezer, line 43: Maybe not a problem of this PR but we should remind people to source their ROS2 installation (which seems to be a common problem among newcomers) `source /opt/ros/foxy/setup.bash`

vatanaksoytezer (approved):
Thanks @tylerjw for taking this. I like the changes overall. A few nits below but approving anyway since I am not sure everyone would agree with my nits.

PR #136, opened 2021-10-10
Remove tutorials that are not yet ported from foxy

tylerjw:
### Description

This removes the not-yet-ported tutorials from the foxy branch.  This will clean up the website to include only tutorials that are ported for ROS 2.


henningkayser (commented):
Shouldn't the tutorials just be commented out or rather than being removed? This way, migrated tutorials would have to added again.

PR #137, opened 2021-10-13
moveit_visual_tools method name was changed

AndyZe:


AndyZe:
Ah, found my problem. Thanks

PR #123, opened 2021-10-01
Add visual tools GUI to MoveGroupInterface Tutorial for demonstration

vatanaksoytezer:
Demonstration for https://github.com/PickNikRobotics/rviz_visual_tools/pull/205. Potentially all the ported tutorials should move back to using rviz visual tools GUI

PR #138, opened 2021-10-13
Update planning_scene_ros_api_tutorial to use rviz prompt

tylerjw:
### Description

This also needs to be done for the main branch too.


vatanaksoytezer (approved):
We can start to use moveit_visual_tools to be consistent with ROS1 tutorials as in https://github.com/ros-planning/moveit2_tutorials/pull/123/files. Though we don't use any mvt features here. I am going to leave the choice to you but lgtm, so approving.

JafarAbdi (approved):

PR #139, opened 2021-10-13
Update planning_scene_ros_api_tutorial to use rviz prompt

tylerjw:
### Description

This is the same as this: https://github.com/ros-planning/moveit2_tutorials/pull/138


vatanaksoytezer (approved):
Also similar comment on #138, we can start to use moveit_visual_tools to be consistent with ROS1 tutorials as in https://github.com/ros-planning/moveit2_tutorials/pull/123/files. Though we don't use any mvt features here. I am going to leave the choice to you but lgtm, so approving.

PR #152, opened 2021-10-18
Fix code links on foxy branch

tylerjw:
### Description

As the API documentation for MoveIt2 doesn't exist yet this PR changes those links to point to source code on the foxy branch.  This also fixes existing links to code to point to the foxy branch.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -20,4 +20,4 @@ jobs:
     - uses: actions/setup-python@v2
     - name: Install clang-format-10
       run: sudo apt-get install clang-format-10
-    - uses: pre-commit/action@v2.0.0
+    - uses: pre-commit/action@v2.0.3


conf.py
```diff
@@ -55,7 +54,7 @@
 # Add any paths that contain custom themes here, relative to this directory.
 
 # Links
-ros1_distro = "noetic"
+ros1_distro = "foxy"
```
auman66, line 57: Should we change this to `ros2_distro` and replace it everywhere in tutorials (can be a separate issue)

auman66 (commented):
I think these changes will add a lot of context.

Questions before I approve: Does CI still run the `htmlproofer.sh` script? If so, then this is good to go. You are introducing a lot of links and this would ensure they all work.

doc/move_group_interface/move_group_interface_tutorial.rst
```diff
@@ -3,7 +3,7 @@ Move Group C++ Interface
...
    :width: 700px
 
-In MoveIt, the simplest user interface is through the :planning_interface:`MoveGroupInterface` class. It provides easy to use functionality for most operations that a user may want to carry out, specifically setting joint or pose goals, creating motion plans, moving the robot, adding objects into the environment and attaching/detaching objects from the robot. This interface communicates over ROS topics, services, and actions to the `MoveGroup Node <http://docs.ros.org/noetic/api/moveit_ros_move_group/html/annotated.html>`_.
+In MoveIt, the simplest user interface is through the :moveit_codedir:`MoveGroupInterface<moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h>` class. It provides easy to use functionality for most operations that a user may want to carry out, specifically setting joint or pose goals, creating motion plans, moving the robot, adding objects into the environment and attaching/detaching objects from the robot. This interface communicates over ROS topics, services, and actions to the :moveit_codedir:`MoveGroup Node <moveit_ros/move_group/src/move_group.cpp>`.
```
henningkayser, line 6:
```suggestion
In MoveIt, the simplest user interface is through the :moveit_codedir:`MoveGroupInterface<moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h>` class. It provides easy to use functionality for most operations that a user may want to carry out, specifically setting joint or pose goals, creating motion plans, moving the robot, adding objects into the environment and attaching/detaching objects from the robot. This interface communicates over ROS topics, services, and actions to the :moveit_codedir:`MoveGroup<moveit_ros/move_group/src/move_group.cpp>` node.
```

conf.py
henningkayser: I agree, or simply ros_distro

henningkayser (approved):

PR #166, opened 2021-11-01
Fix link to MoveGroupInterface

tylerjw:
### Description

This fixes a link to the MoveGroupInterface to instead point to the header file in foxy.

### Checklist


vatanaksoytezer (approved):
I am not sure giving a specific commit is a good idea, but since we don't have docs generated I think this should stay the case.

PR #175, opened 2021-11-02
Do not run Rolling CI on this Galactic branch

AndyZe:
Similar to #174.


vatanaksoytezer (changes requested):
One small thing: You need to change lines 13 and 16 from `main` to `galactic` so we are running CI on this branch as well.

AndyZe:
It looks like the Galactic jobs still are not running, though

vatanaksoytezer:
image

It's running all good.

vatanaksoytezer (approved):

PR #181, opened 2021-11-02
Add mergify for backporting PRs

vatanaksoytezer:
@henningkayser and I talked recently about trying out mergify for backporting (which is already enabled on ros-planning organization), this PR adds mergify to main branch, then we can use mergify to backport mergify in the other branches 😆. For that can we have backport-galactic and backport-foxy labels for that @henningkayser?


henningkayser (approved):

henningkayser:
mergify only had permissions for nav2 :man_facepalming: I just fixed that

PR #201, opened 2021-11-08
Added graph-msgs to vcstool import

### LINKED ISSUE #188: Graph_msgs missing from tutorials
### Description
When running this command:
```rosdep install -r --from-paths . --ignore-src --rosdistro $ROS_DISTRO -y```

Got the following error:

```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
moveit2_tutorials: Cannot locate rosdep definition for [gripper_controllers]
moveit_visual_tools: Cannot locate rosdep definition for [graph_msgs]
```

stephanie-eng:
### Description

Fixes #188 on main branch

### DIFF
--- a/moveit2_tutorials.repos
+++ b/moveit2_tutorials.repos
@@ -11,3 +11,8 @@ repositories:
     type: git
     url: https://github.com/ros-planning/moveit_visual_tools
     version: ros2
+  graph_msgs:
+    type: git
+    url: https://github.com/PickNikRobotics/graph_msgs
+    version: ros2
+    


vatanaksoytezer:
Can you check my comment on #188?

vatanaksoytezer:
Ok so graph_msgs exists on ros-testing but not on ros-main and that's a very good catch! The sync is due today though: https://discourse.ros.org/t/preparing-for-galactic-sync-2021-11-09/22949. So this should be not an issue after today. I would advise not to merge this PR as this would be auto-fixed. Also @tylerjw @rhaschke I think we should make moveit2 docker images depending on main repo (by using *-ci dockers to build upon instead of *-ci-testing) to avoid further issues like this.

stephanie-eng:
I can now see ros-galactic-graph-msgs as a package in apt, so the issue is resolved.

PR #215, opened 2021-11-18
Remove extra / in version

### LINKED ISSUE #214: External links have two slashes
### Description
Minor issue, the external links are not broken, but they all have two slashes in them after the ros distro, for example, the link to move_group_interface from the first link on the Move Group C++ Interface page is:

https://github.com/ros-planning/moveit2/blob/foxy//moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h

The link still works regardless.

stephanie-eng:
### Description

Fixes #214


tylerjw (commented):
That slash is not extra.  Without it a bunch of links get broken.  @vatanaksoytezer found this.  If you have some other way of adding the slash somewhere else in the doc please let me know.

stephanie-eng:
Closing, see #217

PR #217, opened 2021-11-19
Fix extra slashes in github urls

### LINKED ISSUE #214: External links have two slashes

vatanaksoytezer:
Fixes #214.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -174,7 +174,7 @@ extlinks = {
         + html_context["github_user"]
         + "/moveit2_tutorials/blob/"
         + html_context["github_version"]
-        + "/doc/%s",
+        + "doc/%s",
         "",
     ),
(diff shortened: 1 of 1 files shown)


vatanaksoytezer:
I think I found a way to fix that extra slashes, can you test and confirm @tylerjw @stephanie-eng?

stephanie-eng (approved):
Tried to build locally and the extra slash is gone - thanks!

PR #223, opened 2021-11-22
Update servo tutorial for MoveIt servo parameters update

stephanie-eng:
### Description

The [update](https://github.com/ros-planning/moveit2/pull/799) to servo parameters changes the function signature of makeServoParameters. This PR is to update the tutorials accordingly once the aforementioned PR is merged.


stephanie-eng:
The tutorial documentation says to use the files in the moveit package rather than the tutorials, so I'm not sure what servo_cpp_interface_demo.cpp is used for, but it won't build without the change for the servo parameter update.

tylerjw:
I updated this by adding moveit2 to the repos file so this will build against moveit from main.  This should allow this to pass.

AndyZe (approved):

PR #267, opened 2022-01-14
:fire: Fix embeded video

tylerjw:
### Description

Due to the way multiversion deployment works, we have to use relative paths for videos with the raw html feature.  An example of a broken video is in here:

https://moveit.picknik.ai/main/doc/examples/realtime_servo/realtime_servo_tutorial.html#id1

There are probably more clever ways of doing this but I was not able to figure any of them out.


rhaschke (approved):

rhaschke:
Due to multi-version deployment, we have to use relative paths for videos

PR #288, opened 2022-01-14
Add concepts page

jliukkonen:
### Description

This PR adds the Concepts page from [https://moveit.ros.org/documentation/concepts/](https://moveit.ros.org/documentation/concepts/) as part of MoveIt2 documentation. The concepts page should be used as basis for future efforts to improve documentation about high level concepts and important building blocks of MoveIt2.

Note, I added an improved graphic for the PlanningSceneMonitor page. The old one wasn't very useful IMO.


doc/concepts/concepts.rst
```diff
@@ -1,13 +1,228 @@
...
+The figure above shows the high-level system architecture for the primary node provided by MoveIt called [move_group](http://docs.ros.org/noetic/api/moveit_ros_move_group/html/move__group_8cpp.html). This node serves as an integrator: pulling all the individual components together to provide a set of ROS actions and services for users to use.
+
+.. image:: /_static/images/move_group.png
+
+User Interface
+--------------
+
+The users can access the actions and services provided by ``move_group`` in three ways:
+
+- **In C++** - using the `move_group_interface <http://docs.ros.org/noetic/api/moveit_ros_planning_interface/html/classmoveit_1_1planning__interface_1_1MoveGroupInterface.html>`_ package that provides an easy to setup C++ interface to move_group
```
mamoll, line 22: This link should be updated to point to rolling (or, even better, `${moveit_release}`, if Sphinx has some way to expand variables).

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,228 @@
...
+
+- **In Python** - using the `moveit_commander <http://docs.ros.org/noetic/api/moveit_commander/html/classmoveit__commander_1_1move__group_1_1MoveGroupCommander.html>`_ package
```
mamoll, line 24: moveit_commander doesn't exist for MoveIt 2. At the risk of stating the obvious, you could say that in Python (and Rust, Lisp, etc.,) you can interact with MoveIt at the message level.

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,228 @@
...
+
+- **Through a GUI** - using the `Motion Planning plugin to Rviz <https://moveit.picknik.ai/foxy/doc/quickstart_in_rviz/quickstart_in_rviz_tutorial.html>`_ (the ROS visualizer)
```
mamoll, line 26:
```suggestion
- **Through a GUI** - using the :doc:`Motion Planning plugin to Rviz </doc/quickstart_in_rviz/quickstart_in_rviz_tutorial.html>` (the ROS visualizer)
```

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,228 @@
...
+**********
+Kinematics
+**********
+
+The Kinematics Plugin
+---------------------
+
+MoveIt uses a plugin infrastructure, especially targeted towards allowing users to write their own inverse kinematics algorithms. Forward kinematics and finding jacobians is integrated within the RobotState class itself. The default inverse kinematics plugin for MoveIt is configured using the [KDL](https://github.com/orocos/orocos_kinematics_dynamics) numerical jacobian-based solver. This plugin is automatically configured by the MoveIt Setup Assistant.
+
+IKFast Plugin
```
mamoll, line 193: IKFast hasn't been ported yet. Can you comment this out for now?

mamoll (changes requested):
Can you also add the source files for the figures? We will likely update these in the future and it'd nice not to have to start from scratch.

doc/concepts/concepts.rst
jliukkonen: Absolutely. Didn't realize that was the case.

doc/concepts/concepts.rst
jliukkonen: I'll update this to point to the `rolling` API docs once they are online. There's nothing better than `noetic` available for now.

doc/concepts/concepts.rst
jliukkonen: I think I'll just remove the Python entry then. :+1:

doc/concepts/how_to_guide.rst
```diff
@@ -1,43 +0,0 @@
-How-To Guide
-============
-
-The How-To Guides are distinctly different from tutorials in that they are directed towards a different audience.
-How-To Guides are **Problem-Oriented** and assume the reader came to this page to answer a specific question.
-These pages posses these qualities:
-
-* a series of steps
-* a focus on the goal
-* addressing a specific question
```
tylerjw, line 10: you deleted this file instead of moving it, did you mean to do that?

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,228 @@
+========
```
tylerjw, line 1: Would it be useful to break some of this up into separate sub-pages in the concepts section?  That'll make it easier to expand this in the future.

jliukkonen: Yeah, that makes sense to me.

doc/concepts/how_to_guide.rst
jliukkonen: I deleted this deliberately. I didn't quite see a How-To Guide was a MoveIt concept so maybe it should be elsewhere.

doc/concepts/concepts.rst
jliukkonen: The suggested change doesn't seem to work... I'm trying to figure out how to do relative links in RST if you they are supported.

jliukkonen: Apparently like this: 
```doc:`Motion Planning plugin to Rviz <../tutorials/quickstart_in_rviz/quickstart_in_rviz_tutorial>` ```

File: doc/concepts/how_to_guide.rst (hunk 5 above)
Comment (tylerjw) [reply doc/concepts/how_to_guide.rst:10]: Can you figure out where it would fit instead of deleting it?

Comment (tylerjw) [reply doc/concepts/how_to_guide.rst:10]: Maybe create a Concepts or Background section at the top of the corresponding page in Contributing and move it there?

File: doc/concepts/concepts.rst (hunk 1 above)
Comment (tylerjw) [reply doc/concepts/concepts.rst:22]: There is a way.  You use the variable `{DISTRO}`.  But as @jliukkonen said they don't exist yet so this is fine for now.

File: doc/concepts/concepts.rst [hunk 7]
```diff
@@ -1,13 +1,227 @@
...
+
+Robot Interface
+---------------
+
+``move_group`` talks to the robot through ROS topics and actions. It communicates with the robot to get current state information (positions of the joints, etc.), to get point clouds and other sensor data from the robot sensors and to talk to the controllers on the robot.
+
+Joint State Information
+-----------------------
+
+``move_group`` listens on the ``/joint_states`` topic for determining the current state information - i.e. determining where each joint of the robot is. ``move_group`` is capable of listening to multiple publishers on this topic even if they are publishing only partial information about the robot state (e.g. separate publishers may be used for the arm and mobile base of a robot). Note that ``move_group`` will not setup its own joint state publisher - this is something that has to be implemented on each robot.
```
nbbrooks, line 46:
```suggestion
``move_group`` listens on the ``/joint_states`` topic for determining the current state information - i.e. determining where each joint of the robot is. ``move_group`` is capable of listening to multiple publishers on this topic even if they are publishing only partial information about the robot state (e.g. separate publishers may be used for the arm and mobile base of a robot). Note that ``move_group`` will not set up its own joint state publisher - this is something that has to be implemented on each robot.
```

nit - this typo is all over all our docs :(

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,227 @@
...
+
+.. raw:: html
+
+   <hr>
+
+**********************
+Planning Scene Monitor
+**********************
+
+The ``planning scene`` is an object used for storing the representation of the world around the robot and also the state of the robot itself. The internal state of the ``planning_scene`` object is typically maintained by ``planning_scene_monitor`` component that enables reading and writing the state in a thread-safe manner.
```
nbbrooks, line 151:
```suggestion
The ``planning scene`` is an object used for storing the representation of the world around the robot and also the state of the robot itself. The internal state of the ``planning_scene`` object is typically maintained by a ``planning_scene_monitor`` component that enables reading and writing the state in a thread-safe manner.
```

nit - just running this file through a spell/grammar checker

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,227 @@
...
+- **Meshes** - you can use either ``.stl`` (standard triangle language) or ``.dae`` (digital asset exchange) formats to describe objects such as robot links.
+
+- **Primitive Shapes** - e.g. boxes, cylinders, cones, spheres and planes
+
+- **Octomap** - the ``Octomap`` object can be directly used for collision checking
+
+Allowed Collision Matrix (ACM)
+------------------------------
+
+Collision checking is a very expensive operation often accounting for close to 90% of the computational expense during motion planning. The ``Allowed Collision Matrix`` or ``ACM`` encodes a binary value corresponding to the need to check for collision between pairs of bodies (which could be on the robot or in the world). If the value corresponding to two bodies is set to ``true`` in the ACM, it specifies that a collision check between the two bodies is either not required or wanted. The collision checking would not be required if, e.g., the two bodies are always so far way that they can never collide with each other. Alternatively, the two bodies could be in contact with each other by default, in which case the collision detection should be disabled for the pair in the ACM.
```
nbbrooks, line 214:
```suggestion
Collision checking is a very expensive operation often accounting for close to 90% of the computational expense during motion planning. The ``Allowed Collision Matrix`` or ``ACM`` encodes a binary value corresponding to the need to check for collision between pairs of bodies (which could be on the robot or in the world). If the value corresponding to two bodies is set to ``true`` in the ACM, it specifies that a collision check between the two bodies is either not required or wanted. The collision checking would not be required if, e.g., the two bodies are always so far away that they can never collide with each other. Alternatively, the two bodies could be in contact with each other by default, in which case the collision detection should be disabled for the pair in the ACM.
```

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,227 @@
...
+The Kinematics Plugin
+---------------------
+
+MoveIt uses a plugin infrastructure, especially targeted towards allowing users to write their own inverse kinematics algorithms. Forward kinematics and finding jacobians is integrated within the RobotState class itself. The default inverse kinematics plugin for MoveIt is configured using the `KDL <https://github.com/orocos/orocos_kinematics_dynamics>`_ numerical jacobian-based solver. This plugin is automatically configured by the MoveIt Setup Assistant.
+
+******************
+Collision Checking
+******************
+
+Collision checking in MoveIt is configured inside a Planning Scene using the _CollisionWorld_ object. Fortunately, MoveIt is setup so that users never really have to worry about how collision checking is happening. Collision checking in MoveIt is mainly carried out using the `FCL <https://flexible-collision-library.github.io/>`_ package - the primary collision checking library of MoveIt.
```
nbbrooks, line 198:
```suggestion
Collision checking in MoveIt is configured inside a Planning Scene using the _CollisionWorld_ object. Fortunately, MoveIt is set up so that users never really have to worry about how collision checking is happening. Collision checking in MoveIt is mainly carried out using the `FCL <https://flexible-collision-library.github.io/>`_ package - the primary collision checking library of MoveIt.
```

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,227 @@
...
+***********************
+
+The figure above shows the high-level system architecture for the primary node provided by MoveIt called `move_group <http://docs.ros.org/noetic/api/moveit_ros_move_group/html/move__group_8cpp.html>`_. This node serves as an integrator: pulling all the individual components together to provide a set of ROS actions and services for users to use.
+
+.. image:: /_static/images/move_group.png
+
+User Interface
+--------------
+
+The users can access the actions and services provided by ``move_group`` in three ways:
```
nbbrooks, line 20:
```suggestion
The users can access the actions and services provided by ``move_group`` in two ways:
```

commander isn't available, we'll have to get around to updating diagram later

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,227 @@
...
-These are the **why-docs** that discuss the design of MoveIt and this website.
+The following list major MoveIt concepts and components and describes how they tie in with each other. For tutorials, see the `tutorials <https://moveit.picknik.ai/>`_ page.
+
+.. image:: /_static/images/moveit_pipeline.png
+
+***********************
+The ``move_group`` node
+***********************
+
+The figure above shows the high-level system architecture for the primary node provided by MoveIt called `move_group <http://docs.ros.org/noetic/api/moveit_ros_move_group/html/move__group_8cpp.html>`_. This node serves as an integrator: pulling all the individual components together to provide a set of ROS actions and services for users to use.
```
nbbrooks, line 13:
```suggestion
The figure below shows the high-level system architecture for a key node provided by MoveIt called `move_group <https://docs.ros.org/noetic/api/moveit_ros_move_group/html/move__group_8cpp.html>`_. This node serves as an integrator: pulling all the individual components together to provide a set of ROS actions and services for users to use. Note that moveit_commander is `not currently available in MoveIt 2 <https://github.com/ros-planning/moveit2/issues/314>`_.
```

moveit cpp probably didn't exist when this was written

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,227 @@
...
+- **Point clouds**: handled by the ``point cloud occupancy map updater`` plugin
+
+- **Depth images**: handled by the ``depth image occupancy map updater`` plugin
+
+Note that you can add your own types of updaters as a plugin to the occupancy map monitor.
+
+Octomap
+-------
+
+The Occupancy map monitor uses an `Octomap <http://octomap.github.io/>`_ to maintain the occupancy map of the environment. The Octomap can actually encode probabilistic information about individual cells although this information is not currently used in MoveIt. The Octomap can directly be passed into FCL, the collision checking library that MoveIt uses.
```
nbbrooks, line 174:
```suggestion
The Occupancy map monitor uses an `Octomap <https://octomap.github.io/>`_ to maintain the occupancy map of the environment. The Octomap can actually encode probabilistic information about individual cells although this information is not currently used in MoveIt. The Octomap can directly be passed into FCL, the collision checking library that MoveIt uses.
```

doc/concepts/concepts.rst
```diff
@@ -1,13 +1,227 @@
...
+*********************
 
-.. toctree::
-   :maxdepth: 1
+Time parameterization
+---------------------
 
-   tutorials
-   how_to_guide
+Motion planners typically only generate “paths”, i.e. there is no timing information associated with the paths. MoveIt includes several `trajectory processing algorithms <http://docs.ros.org/en/noetic/api/moveit_core/html/cpp/classtrajectory__processing_1_1TimeOptimalTrajectoryGeneration.html>`_ that can work on these paths and generate trajectories that are properly time-parameterized accounting for the maximum velocity and acceleration limits imposed on individual joints. These limits are read from a special ``joint_limits.yaml`` configuration file that is specified for each robot. The configuration file is optional and it overrides any velocity or acceleration limits from the URDF. The recommended algorithm as of January 2022 is **time_optimal_trajectory_generation** (TOTG). A caveat for this algorithm is that the robot must start and end at rest. By default, the TOTG timestep is 0.1 seconds.
```
nbbrooks, line 227:
```suggestion
Motion planners typically only generate “paths”, i.e. there is no timing information associated with the paths. MoveIt includes several `trajectory processing algorithms <https://docs.ros.org/en/noetic/api/moveit_core/html/cpp/classtrajectory__processing_1_1TimeOptimalTrajectoryGeneration.html>`_ that can work on these paths and generate trajectories that are properly time-parameterized accounting for the maximum velocity and acceleration limits imposed on individual joints. These limits are read from a special ``joint_limits.yaml`` configuration file that is specified for each robot. The configuration file is optional and it overrides any velocity or acceleration limits from the URDF. The recommended algorithm as of January 2022 is **time_optimal_trajectory_generation** (TOTG). A caveat for this algorithm is that the robot must start and end at rest. By default, the TOTG timestep is 0.1 seconds.
```

nbbrooks (approved):
some small nits, let's get this merged in so we can begin larger changes!

tylerjw:
Nit. It makes it much easier try track changes when each sentence ends in a newline.  rst creates a line break only after an empty line for this reason.

doc/concepts/concepts.rst
jliukkonen: Yeah, I think moveitcpp didn't exist back then.

mamoll (approved):

PR #298, opened 2022-01-15
:memo: Make Galactic branch the default

tylerjw:
### Description

This will make the galactic branch of this website the default landing page.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ help:
 
 multiversion: Makefile
 	sphinx-multiversion $(OPTS) "$(SOURCE)" build/html
-	@echo "<html><head><meta http-equiv=\"refresh\" content=\"0; url=foxy/index.html\" /></head></html>" > build/html/index.html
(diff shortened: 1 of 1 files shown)


AndyZe:
Let's take a moment and look at David Lu's distro starts before merging this. I'll do it soon .

tylerjw:
This one? https://metrics.ros.org/rosdistro_rosdistro.html

There are more users than foxy or rolling per that.

AndyZe:
This graph shows Foxy bring about 2X more popular than Galactic.

https://metrics.ros.org/packages_rosdistro.html

Meh, whatever.

AndyZe (approved):

tylerjw:
Do you actually want to encourage people to use foxy?  We are library developers.  Do you want MoveIt 2 to be based on foxy?

AndyZe:
My point is, Galactic isn't used much. It probably never will be because it's non-LTS. That's all.

AndyZe:
It's fine. A 2X factor is actually better than I expected.

PR #304, opened 2022-01-15
Add sphinx dependecies into requirements

destogl:
### Description

Adding sphinx and its plugins as dependencies.

### DIFF
--- a/doc/how_to_contribute/how_to_contribute_to_site.rst
+++ b/doc/how_to_contribute/how_to_contribute_to_site.rst
@@ -16,6 +16,10 @@ Requirements
 - ROS 2 Galactic
 - Docker
 - A colcon workspace with a copy of `the moveit2_tutorials repo <https://github.com/ros-planning/moveit2_tutorials>`_ (if you don't already have one, :doc:`/doc/tutorials/getting_started/getting_started` will walk you through the process of creating one).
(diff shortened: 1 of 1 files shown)


doc/how_to_contribute/how_to_contribute_to_site.rst
```diff
@@ -16,6 +16,10 @@ Requirements
 - ROS 2 Galactic
 - Docker
 - A colcon workspace with a copy of `the moveit2_tutorials repo <https://github.com/ros-planning/moveit2_tutorials>`_ (if you don't already have one, :doc:`/doc/tutorials/getting_started/getting_started` will walk you through the process of creating one).
+- Sphinx pages generator with `multitab`, `multiversion`, and `copybutton` extensions:
+  ```
+  pip3 install sphinx-multiversion sphinx-tabs sphinx-copybutton  # use `sudo` for system-wide installation
```
tylerjw, line 21: Shouldn't we just use the requirements.txt file here: https://github.com/ros-planning/moveit2_tutorials/blob/main/requirements.txt

Secondly, pip is a weird animal, and to be as nice as possible to people's system we should invoke it like this:

```
python3 -m pip install --user --upgrade --requirement requirements.txt
```

PR #328, opened 2022-02-07
Add Hybrid Planning concepts page

henningkayser:
Adds the content of the Hybrid Planning white paper as concepts page.


doc/concepts/hybrid_planning/hybrid_planning.rst
```diff
@@ -0,0 +1,171 @@
+===============
+Hybrid Planning
+===============
+
+MoveIt’s motion planning architecture follows the “Sense-Plan-Act” approach. In order for motions to get planned and executed, the environment and the robot state are first perceived (“Sense”), then the robot trajectory is computed by the planner (“Plan”) and finally executed in a single run using a trajectory controller (“Act”).
```
AndyZe, line 5:
```suggestion
MoveIt’s motion planning architecture previously followed the “Sense-Plan-Act” approach. In order for motions to get planned and executed, the environment and the robot state are first perceived (“Sense”), then the robot trajectory is computed by the planner (“Plan”) and finally executed in a single run using a trajectory controller (“Act”).
```

doc/concepts/hybrid_planning/hybrid_planning.rst
```diff
@@ -0,0 +1,171 @@
...
+
+While this solution works well for global motion planning in well-known static environments, the approach is not applicable in many real-world applications, in particular in unstable or dynamic environments. Tasks like serving a glass of water to a person on a table or writing on an uneven chalkboard require more sophisticated methods that allow reacting to unpredictable changes. For instance, the environment of the robot could change dynamically, i.e. the serviced person may move so that the robot has to update the hand-over posture. Or there could be certain uncertainties in the task itself, i.e. writing with chalk requires adapting the pressure on the board while the chalk also gets shorter by being used up.
```
AndyZe, line 7:
```suggestion
While this solution works well for global motion planning in well-known static environments, the approach is not applicable in many real-world applications, in particular in unstable or dynamic environments. Tasks like serving a glass of water to a person on a table or writing on an uneven chalkboard require more sophisticated methods that allow reacting to unpredictable changes. For instance, the environment of the robot could change dynamically, or there could be certain uncertainties in the task itself, e.g. writing with chalk requires adapting the pressure on the board while the chalk also gets shorter by being used up. 
```

tylerjw (commented):
Thank you for writing this.  This helped answer some questions I had about how the internal communication of hybrid planner worked.

doc/concepts/hybrid_planning/hybrid_planning.rst
henningkayser: I think it still does, adding "previously" might be misleading as people could think that this automagically works with the planning pipeline.

tylerjw (approved):

PR #345, opened 2022-03-30
Fix htmlproofer and Rolling / Jammy CI

vatanaksoytezer:
Still need to update instructions for 22.04 and Rolling.


moveit2_tutorials.repos
```diff
@@ -1,12 +1,8 @@
 repositories:
-  moveit2:
-    type: git
-    url: https://github.com/ros-planning/moveit2
-    version: main
```
vatanaksoytezer, line 5: Remove this because this is redundant for CI, instead I added it to instructions.

tylerjw (approved):

PR #353, opened 2022-05-05
[galactic] Update black version

stephanie-eng:
Should fix the error in formatting CI - see https://github.com/ros-planning/moveit2_tutorials/pull/344

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
     hooks:
       - id: black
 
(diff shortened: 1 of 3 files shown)


AndyZe (approved):

AndyZe:
I was actually looking at this earlier. I think you need to add the github permissions like this:

https://github.com/ros-planning/moveit2/pull/1210

stephanie-eng:
> I was actually looking at this earlier. I think you need to add the github permissions like this:
> 
> [ros-planning/moveit2#1210](https://github.com/ros-planning/moveit2/pull/1210)

I saw this error being commented about [here](https://github.com/actions/checkout/issues/760), which should be fixed with a new version of checkout, hopefully. I just made a new PR #354 to see if that works.

rhaschke (approved):
Although this PR doesn't solve the "git safe directory" issue (not sure this was intended?),
it actually solves issues with black auto-formatting. Thus I approve.
@stephanie-eng, please be more specific in your PR description.

stephanie-eng:
Apologies, see updated description. This was not intended to fix the git safe directory issue, just the error from black in the formatting check.

PR #355, opened 2022-05-05
Fix CI permission errors

stephanie-eng:
### Description

As per https://github.com/ros-planning/moveit2/pull/1206


rhaschke (commented):
Changing the git config at the level of GHA will not work.
You need to change it within the docker container.
https://github.com/ros-planning/moveit2/pull/1215 should solve the issue in general.

PR #357, opened 2022-05-06
Update launch file to use output="both"

AndyZe:
### Description

Fix this error due to a recent ROS2 Rolling change:

`stdoutstderr is not a valid standard output config i.e. "screen", "log" or "both"`


JafarAbdi (approved):

JafarAbdi:
The failure isn't related to this PR

PR #359, opened 2022-05-10
Fixup main branch

tylerjw:
Signed-off-by: Tyler Weaver <tylerjw@gmail.com>

### Description

This fixes up the main branch so it builds on Jammy

### Checklist

### DIFF
--- a/moveit2_tutorials.repos
+++ b/moveit2_tutorials.repos
@@ -1,21 +1,29 @@
 repositories:
+  launch_param_builder:
+    type: git
+    url: https://github.com/PickNikRobotics/launch_param_builder
+    version: main
+  moveit2:
+    type: git
+    url: https://github.com/ros-planning/moveit2
+    version: main
   moveit_resources:
     type: git
     url: https://github.com/ros-planning/moveit_resources
+    version: ros2
+  moveit_task_constructor:
+    type: git
+    url: https://github.com/ros-planning/moveit_task_constructor.git
     version: ros2
   moveit_visual_tools:
     type: git
     url: https://github.com/ros-planning/moveit_visual_tools
     version: ros2
-  launch_param_builder:
-    type: git
-    url: https://github.com/PickNikRobotics/launch_param_builder
-    version: main
-  moveit_task_constructor:
-    type: git
-    url: https://github.com/ros-planning/moveit_task_constructor.git
-    version: ros2
   rosparam_shortcuts:
     type: git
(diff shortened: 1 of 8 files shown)


moveit2_tutorials.repos
```diff
@@ -1,21 +1,29 @@
...
-    version: ros2
-  moveit_visual_tools:
-    type: git
-    url: https://github.com/ros-planning/moveit_visual_tools
-    version: ros2
   launch_param_builder:
     type: git
     url: https://github.com/PickNikRobotics/launch_param_builder
     version: main
+  moveit2:
```
vatanaksoytezer, line 6: We are already using the latest moveit2 via the docker container. This is redundant.

vatanaksoytezer (commented):
Small note below, rest looks good, thank you Tyler!

henningkayser (approved):

moveit2_tutorials.repos
vatanaksoytezer: So there is difference between our CI and getting started instructions. CI utilizes the pre-built docker container of moveit2 and sources the pre-built workspace (see https://github.com/ros-planning/moveit2_tutorials/blob/main/.github/workflows/ci.yaml#L29), while the getting started utilizes the source built into people's local machine, that was the best way I found maintaining it, but I'm happy to change it. If we add moveit2 in .repos file we would redundantly build moveit2 in CI which would take a significant time.

stephanie-eng: OK - there's a clear reason to remove it then. The tutorial text as-is says to clone it anyway, so in that case there's no need to update.

stephanie-eng (approved):
Looks good pending the bit about moveit2 in the .repos file

moveit2_tutorials.repos
tylerjw: What do you think of having a separate repo file for CI for this project? I think that could make this easier in a way because it would be automated which repos we test with while keeping the user instructions as simple as possible.  Note that I added more to this file than just moveit2 as the current instructions don't work without also adding srdfdom.

vatanaksoytezer: I'm happy with that as long as we make it happen either in this PR or a quick follow up PR.

vatanaksoytezer: We should update the getting started accordingly as well.

tylerjw: I'll do that here.

PR #368, opened 2022-05-19
Update ompl->chomp instructions

AndyZe:
### Description

Update instructions for post-processing OMPL output with CHOMP.

I don't have time to update the rest of the CHOMP tutorial, but I do want to update the usage here since it took me awhile to figure out.


davetcoleman (approved):
I read through it and it looks right to me

PR #371, opened 2022-05-21
Add hints on debugging MTC from terminal output

AndyZe:
Also delete some unused images I happened to notice.


tylerjw (approved):
Thank you for this.

Issue #404, opened 2022-06-02
Why Do Humble and Rolling Tutorials Require Building MoveIt From Source?


davetcoleman:
In the ``moveit2_tutorials.repos`` it lists moveit.git as a dependency. I can imagine at one point something was out of sync from the debians so this was required. However should this be standard? It means our users must spend a lot of time building.

I don't feel strongly either way but the getting started tutorial does not clarify this is happening. It instead says on the rosdep step:
> The following will install from Debian any package dependencies not already in your workspace. This is the step that will install MoveIt and all of its dependencies:

I am working on cleaning up that tutorial to clarify it is being built from source. But am not sure this is the best idea. @tylerjw

mikeferguson:
It took me 56 minutes to build JUST moveit2 and srdfdom (which had a change that I needed to use). It looks like latest debs are not past the test server yet for humble

vatanaksoytezer:
We have released, but we are waiting a sync on Rolling and Humble. Stable branches (humble, galactic, foxy) of tutorials should use debians, main branch should use source since main branch of the tutorials can be updated with a change in main branch of moveit.

vatanaksoytezer:
> I've been trying to build MoveIt, following our getting started tutorial, for over 30 minutes now. It's not been a fun experience.

We actually have prebuilt docker images, that can people use that is documented [here](https://moveit.picknik.ai/humble/doc/how_to_guides/how_to_setup_docker_containers_in_ubuntu.html). In my opinion we should promote it.

davetcoleman:
My colcon build keeps failing silently... I think I'm running out of memory because I'm in a VMware box with 8 GB ram (on a 32 GB laptop). It fails silently because my terminal just closes when it does. I'm not having a great "Getting Started" experience so far... still haven't gotten past this first tutorial.

davetcoleman:
>We actually have prebuilt docker images, that can people use that is documented [here](https://moveit.picknik.ai/humble/doc/how_to_guides/how_to_setup_docker_containers_in_ubuntu.html). In my opinion we should promote it.

I can add this to the Getting Started tutorial now

smartin015:
Hey folks - I love that there's now a Docker approach to getting started with MoveIt, but there's still a couple snags that will trip up users unfamiliar with Docker / Moveit / Ros2 etc. Here's my notes:

```
# After fetching the start-docker script and chmod +x
./start-docker.sh moveit2-humble moveit/moveit2:humble-source
[...]

PR #422, opened 2022-06-07
Replace start-docker.sh with docker-compose.yml

smartin015:
### Description

Per observations/discussion in #404, this PR replaces the original `start-docker.sh` script with a `docker-compose.yml` script. The same features are available in both, but using a compose file is less custom and easier to customize. It uses an anonymous volume - all changes in `/root` are persisted between runs.

Also fixes a couple of typos I found in the tutorials :)

@vatanaksoytezer please give this script a shot and let me know if there's anything you'd like me to change (e.g. if you'd prefer to also keep the script so folks have a fallback)

### Checklist


gavanderhoorn:
Does this still run `xhost` to update the permissions? Or is that no longer needed either?

smartin015:
> Does this still run `xhost` to update the permissions? Or is that no longer needed either?

xhost should no longer be needed as `~/.Xauthority` is mounted in (see https://stackoverflow.com/a/69505647). Please give it a try though - it works on my setup, but I haven't tried on a different PC.

doc/how_to_guides/how_to_setup_docker_containers_in_ubuntu.rst
```diff
@@ -1,68 +1,44 @@
...
+3. Launch the container.
 
-  There are 3 parameters for the script:
-      - ``name_of_the_container`` : this is the name you wish to give the created container. For this guide, we will be naming the container ``moveit2-rolling``.
-      - ``name_of_the_image`` : if you are creating a fresh Docker container, provide the name of the Docker image here. For this guide, we will be using the image ``moveit/moveit2:rolling-source``. Further explanation of this parameter is provided in the ``Further Reading`` section.
-      - ``using_gpu`` : if ``true``, the Docker will be run using Nvidia GPU drivers. By default, this value is true.
+   .. code-block:: bash
 
-  To run the script and use Nvidia GPU drivers
+    ROS_DISTRO=rolling-source docker-compose run gpu
```
vatanaksoytezer, line 35: I had to do this without dash:
```suggestion
    ROS_DISTRO=rolling-source docker compose run gpu
```

vatanaksoytezer: A bit nitpicky but the name `ROS_DISTRO` is a bit confusing here. I would prefer to either one name: DOCKER_IMAGE or some sorts; or have two different parameters, ROS_DISTRO (Which would be rolling, humble, foxy etc.) and BUILD, which would be either source or release.

doc/tutorials/visualizing_in_rviz/visualizing_in_rviz.rst
```diff
@@ -112,7 +112,7 @@ After we've constructed and initialized, we now create some closures (function o
 
 .. code-block:: C++
 
-    // Create a closures for visualization
+    // Create closures for visualization
```
vatanaksoytezer, line 115: A bit nitpicky again: Do you mind changing this and the one below in another PR so it is not out of context?

doc/how_to_guides/how_to_setup_docker_containers_in_ubuntu.rst
vatanaksoytezer: Of course we should change the documentation according to new parameter

doc/tutorials/your_first_project/your_first_project.rst
```diff
@@ -200,7 +200,7 @@ If it fails to find that within 10 seconds, it prints this error and terminates
 
 The first thing we do is create the MoveGroupInterface. This object will be used to interact with move_group, which allows us to plan and execute trajectories.
 Note that this is the only mutable object that we create in this program.
-Another thing to take note of is the second interface to the ``MoveGroupInterface`` object we are creating here: ``"panda_arm"``.
+Another thing to take note of is the second argument to the ``MoveGroupInterface`` object we are creating here: ``"panda_arm"``.
```
vatanaksoytezer, line 203: Same, in another PR

vatanaksoytezer (changes requested):
Hello @smartin015 thank you so much for the PR, this is way nicer than scripting! I did test this and was able to get it to work with a few changes, that I mentioned below. I have a few nitpicky comments, appreciate if you can take them a look, otherwise, it's all good. Thank you so much!

doc/how_to_guides/how_to_setup_docker_containers_in_ubuntu.rst
smartin015: Interesting- apparently `docker compose` without the hyphen is a new thing :) will add a note that the hyphen may still be required for older versions of compose (and change the env variable name)

doc/tutorials/visualizing_in_rviz/visualizing_in_rviz.rst
smartin015: Done

doc/tutorials/your_first_project/your_first_project.rst
smartin015: Done

vatanaksoytezer (approved):

vatanaksoytezer:
Thank you very much @smartin015 for the contribution, much appreciated!

gavanderhoorn:
With the latest `docker-compose-plugin` & `nvidia-docker2` available for Focal, and the `.yaml` file contributed by this PR, I get `could not connect to display :1` errors if I don't first run `xhost +` (or a more secure version of that command).

@smartin015: what sort of system are you running?

_scripts/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      - moveit2-data:/root
+    environment:
+      DISPLAY: $DISPLAY
+  gpu:
+    <<: *base-gpu
+    image: moveit/moveit2:$DOCKER_IMAGE
+    volumes:
+      - moveit2-data:/root
+    environment:
+      DISPLAY: $DISPLAY
```
gavanderhoorn, line 42: @smartin015: I'm wondering if this works as intended.

According to the YAML spec and my testing, this does not override/merge `environment` into the `environment` map specified in `x-base-gpu`, which doesn't override/merge the `environment` map specified in `x-base-cpu`.

I've had to do something like this to get it to work:

```yaml
x-base-cpu: &base-cpu
  ...
  environment: &base-cpu-env
    QT_X11_NO_MITSHM: 1
  ...

x-base-gpu: &base-gpu
  <<: *base-cpu
  ...
  environment: &base-gpu-env
    <<: *base-cpu-env
    NVIDIA_VISIBLE_DEVICES: all
    NVIDIA_DRIVER_CAPABILITIES: all

services:
  ...
  gpu:
    <<: *base-gpu
    ...
    environment:
      <<: *base-gpu-env
      DISPLAY: $DISPLAY
```

could you perhaps verify `QT_X11_NO_MITSHM` is being set in the `gpu` container @smartin015? `env | sort` didn't show it for me, unless I made the above changes.

vatanaksoytezer: My bad on testing somewhat I may have misused some settings while testing. I can confirm what @gavanderhoorn says here. The docker container launches fine but the display isn't working. This fix seems to set the correct environment variables as suggested, but unfortunately, the display wasn't working for me (ie: xeyes). I'm going to revert this change in a day if not resolved to keep non working stuff away from main.

smartin015: Apologies for the breakage - I opened #439 to remove the `x-*` templating since it may be a bit suspect. However, please do revert the change if you'd rather not try to fix forward. Better to have working old instructions than non-working new ones :)

From a clean boot of my Ubuntu 22.04 computer, I ran 

```
DOCKER_IMAGE=humble-source docker-compose run gpu /bin/bash
source install/setup.bash && ros2 run rviz2 rviz2
```

Running `xhost` command shows only `SI:localuser:<my username>`, so it seems that xhost isn't configured in any way beyond the defaults.

PR #426, opened 2022-06-07
Update tutorial to avoid moving the MTC task

stephanie-eng:
### Description

Running the existing MTC tutorial code will segfault with the latest version of MoveIt Task Constructor. Workaround is to avoid moving/copying the MTC task.

### Checklist

### DIFF
--- a/doc/tutorials/pick_and_place_with_moveit_task_constructor/pick_and_place_with_moveit_task_constructor.rst
+++ b/doc/tutorials/pick_and_place_with_moveit_task_constructor/pick_and_place_with_moveit_task_constructor.rst
@@ -201,7 +201,7 @@ Open ``mtc_tutorial.cpp`` in your editor of choice, and paste in the following c
 
     private:
       // Compose an MTC task from a series of stages.
-      mtc::Task createTask();
+      void createTask(mtc::Task& task);
       mtc::Task task_;
       rclcpp::Node::SharedPtr node_;
     };
@@ -236,7 +236,7 @@ Open ``mtc_tutorial.cpp`` in your editor of choice, and paste in the following c
 
     void MTCTaskNode::doTask()
     {
-      task_ = createTask();
+      createTask(task_);
 
       try
       {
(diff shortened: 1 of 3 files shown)


henrygerardmoore (approved):
Cloned and ran and it works as expected now! Didn't encounter any problems

JafarAbdi:
> Obviously, it would be better to fix MTC. Is there an issue for the underlying problem? Could you pinpoint the problem? I think this worked with MoveIt1.

@rhaschke I was debugging this issue yesterday and came up with a minimal reproducible example, I'll open a draft PR later today

```cpp
TEST(Task, taskMoveContructor) {
	rclcpp::init(0, nullptr);
	auto createTask = [] {
		moveit::core::RobotModelConstPtr robot_model = getModel();
		Task t("first");
		t.setRobotModel(robot_model);
		resetMockupIds();
		auto ref = new stages::FixedState("fixed");
		auto scene = std::make_shared<planning_scene::PlanningScene>(t.getRobotModel());
		ref->setState(scene);

		t.add(Stage::pointer(ref));
		t.add(std::make_unique<ConnectMockup>());
		t.add(std::make_unique<MonitoringGeneratorMockup>(ref));
		return t;
	};

	// Doesn't segfault
	/* Task t = createTask(); */
	// Segfaults when introspection is enabled
	Task t;
	t = createTask();

	try {
		t.init();
		EXPECT_TRUE(t.plan(1));
	} catch (const InitStageException& e) {
		ADD_FAILURE() << "InitStageException:" << std::endl << e << t;
	}
}
```

stephanie-eng:
I agree that we should fix MTC as well, but I also think this workaround is a better usage example for the tutorials. It's a bit cleaner to read, and there's really no need to move the task around when we can just pass a reference to it.

rhaschke:
> It's a bit cleaner to read, and there's really no need to move the task around when we can just pass a reference to it.

I don't really agree. I think it is quite natural to have a factory function for task creation instead of "filling" a provided task instance.

tylerjw:
Generally, I think it is not advisable to use mutable reference parameters. I agree that the better option is to fix MTC than to require/encourage people to create mutable references.

stephanie-eng:
OK, then we can leave the tutorial as-is - I'll close this PR once MTC is fixed.

PR #460, opened 2022-07-05
Use 1/0 instead of true/false in ompl_planning.yaml

stephanie-eng:
If you set the parameter in ompl_planning.yaml to "true" or "false" as suggested in the persistent roadmap section, you will get the following error when planning:

```
[moveit.planning_request_adapter]: Exception caught executing adapter 'Fix Start State In Collision': bad lexical cast: source type value could not be interpreted as target
[task_planning_action_server-1] Skipping adapter instead.`
```

In ompl_interface, these parameters are set using `boost::lexical_cast<bool>(it->second)`, which expects numbers as strings ("1" or "0") but will fail if passed a string like "true" or "false".

### DIFF
--- a/doc/examples/ompl_interface/ompl_interface_tutorial.rst
+++ b/doc/examples/ompl_interface/ompl_interface_tutorial.rst
@@ -142,26 +142,26 @@ By default the planning algorithms start from scratch for each motion planning r
 
     PersistentLazyPRMstar: # use this with a representative environment to create a roadmap
         type: geometric::LazyPRMstar
-        multi_query_planning_enabled: true
-        store_planner_data: true
-        load_planner_data: false
+        multi_query_planning_enabled: 1
+        store_planner_data: 1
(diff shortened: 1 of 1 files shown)


AndyZe:
Sorry I forgot about this issue:

https://mail.google.com/mail/u/1/#inbox/FMfcgzGpGKcpFztMGWsSnBDfWZFNTWph

AndyZe:
In my opinion, it would be better to fix this in source code rather than updating the documentation.

stephanie-eng:
It would be nice to be ale to accept true/false, but we might want this change for consistency if anything. None of the other OMPL planning parameters use "true" or "false", they're set using booleans, for example: https://github.com/ros-planning/moveit_resources/blob/2c58cf3a2ed4e878a7a491572133670a44573c98/panda_moveit_config/config/ompl_planning.yaml#L43-L47

AndyZe (approved):

PR #481, opened 2022-08-02
Delete unused config in ompl planning tutorial

AndyZe:
### Description

It looks like this config folder is not used at all. Instead, `moveit_resources_panda` is used.


vatanaksoytezer (commented):
This file is not being used, but removing this would break the link on https://github.com/ros-planning/moveit2_tutorials/blob/bcf9fa886c30180fdad83876508a9e36c7929dc4/doc/how_to_guides/using_ompl_constrained_planning/ompl_constrained_planning.rst, can you redirect that to one on panda as well?

AndyZe:
Ah, I understand better now. That config file was left as an example. I still think it's better to remove it b/c one would expect the config file in the folder to be the one that's used. Done.

(Also fixed a launch typo)

vatanaksoytezer (approved):

PR #486, opened 2022-08-08
Add hints about OMPL roadmap caching

AndyZe:


mamoll (approved):
Thanks for adding that clarification.

PR #507, opened 2022-09-06
Add note about switching middlewares (to CycloneDDS)

Sgt-Harry:
### Description
Execution phase of the Step Visualizing_in_rviz may fail due to the default middlewares. Switching it to CycloneDDS can solve the problem.
I add some notes so more people can see it.
More details are discussed here:[https://github.com/ros-planning/moveit2/issues/1474#issuecomment-1227509627](url)

### DIFF
--- a/doc/tutorials/visualizing_in_rviz/visualizing_in_rviz.rst
+++ b/doc/tutorials/visualizing_in_rviz/visualizing_in_rviz.rst
@@ -270,6 +270,7 @@ To continue, press ``Next`` again to see your robot execute the plan.
 
 .. image:: executing.png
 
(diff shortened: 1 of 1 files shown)


Sgt-Harry:
> @Sgt-Harry I'm sorry I didn't notice this PR until I'd started #519. I think there are some things that could use improvement with this PR before we can merge it:
> 
> * I think we should put the note in `doc/tutorials/getting_started/getting_started.rst`, not in `visualizing_in_rviz.rst`. The former should be seen by more people.
> * I'd also suggest a note in `quickstart_in_rviz_tutorial.rst` because that is the first one most people will see.
> * Can you update the wording to be more similar to #519?
 
Its my first time making a PR on Github. So in many ways I lack experience. If I do something really stupid, hope that wonst bother you too much.

PR #514, opened 2022-09-17
Re-order Executor initialization to prevent a ROS2 bug

AbduEhab:
### Description

I'm specifically talking about this issue: https://github.com/ros-planning/moveit2/issues/1474. I only changed the order in which code is executed by creating the MoveGroup Interface before the executor.

The format of the code was not changed. The text explaining what the code does was also changed to represent the newly added/removed lines.

PR #516, opened 2022-09-19
Reordered code execution

AndyZe:
**Description**

I'm specifically talking about this issue: https://github.com/ros-planning/moveit2/issues/1474. I only changed the order in which code is executed by creating the MoveGroup Interface before the executor.

The format of the code was not changed. The text explaining what the code does was also changed to represent the newly added/removed lines.

Takes over from PR #514


AbduEhab:
Are there other files that need to be modified before we merge the branch? I haven't been through the entire docs to confirm any other cases, but so far this is it.

AndyZe:
We are only waiting on CI to pass. It's nothing related to this PR. Sorry for the delay

AndyZe:
I think the correct fix for this is to enable Cyclone DDS, which is addressed in #519. So I vote to close this PR.

adlarkin:
@AndyZe should we go ahead and close this, since the fix is enabling cyclone DDS which is [mentioned in the tutorials](https://moveit.picknik.ai/humble/doc/tutorials/getting_started/getting_started.html#switch-to-cyclone-dds)?

Issue #511, opened 2022-09-14
moveit_ros_control_interface failed with colcon build --mixin release


enriLoniterp:
### Description

I'm trying to do the basic tutorial of Moveit2 humble. I have followed all the procedures of the Getting Started but `moveit_ros_control_interface` ifailed

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04
* Source binary
* If binary, which release version?
* If source, which git commit or tag?

### Steps to reproduce
https://moveit.picknik.ai/humble/doc/tutorials/getting_started/getting_started.html

Uninstalled ROS 2 Humble, deleted every dir created also in opt.
Reinstalled, and get to this tutorial

### Expected behaviour
It should compile with colcon build --mixin release, instead it gave me error on a packace

### Backtrace or Console output
```
rosdev@rosdev:~/ws_moveit2/src$ source /opt/ros/humble/setup.bash
rosdev@rosdev:~/ws_moveit2/src$ colcon build --mixin release
[0.466s] WARNING:colcon.colcon_core.package_selection:Some selected packages are already built in one or more underlay workspaces:
	'hardware_interface' is in: /opt/ros/humble
	'controller_interface' is in: /opt/ros/humble
If a package in a merged underlay workspace is overridden and it installs headers, then all packages in the overlay must sort their include directories by workspace order. Failure to do so may result in build failures or undefined behavior at run time.
If the overridden package is used by another package in any underlay, then the overriding package in the overlay must be API and ABI compatible or undefined behavior at run time may occur.

If you understand the risks and want to override a package anyways, add the following to the command line:
	--allow-overriding controller_interface hardware_interface

This may be promoted to an error in a future release of colcon-override-check.
Starting >>> moveit_common
Starting >>> moveit_resources_panda_description
Starting >>> moveit_configs_utils
Starting >>> moveit_resources_fanuc_description
Starting >>> controller_manager_msgs
Starting >>> ros2_control_test_assets
Starting >>> moveit_task_constructor_msgs
Starting >>> rviz_marker_tools
Finished <<< moveit_common [1.09s]                                   
Starting >>> moveit_resources_pr2_description
Finished <<< moveit_configs_utils [1.10s]
Finished <<< moveit_resources_panda_description [1.12s]
Starting >>> moveit_resources_prbt_support
Starting >>> moveit_resources_panda_moveit_config
Finished <<< moveit_resources_fanuc_description [1.18s]
[...]

PR #517, opened 2022-09-20
Update ros2_control version to pull in chained controllers information

eholum:
### Description

Addresses https://github.com/ros-planning/moveit2_tutorials/issues/511

This change to support chained controllers https://github.com/ros-planning/moveit2/pull/1482 relies on the updated msg format from [ros2_control](https://github.com/ros-controls/ros2_control/commit/94d1e9a9f07c92094992d26c44ac8d9e0a3c9103#diff-ff07c5d9557e55b54b7db28eb20db07c45d7b1f340a784dd1178d704eed2776b).

Otherwise we end up with build errors,

```
ws_moveit2/src/moveit2/moveit_plugins/moveit_ros_control_interface/src/controller_manager_plugin.cpp:471:22: error: ‘struct controller_manager_msgs::msg::ControllerState_<std::allocator<void> >’ has no member named ‘chain_connections’
  471 |       if (controller.chain_connections.size() > 1)
      |                      ^~~~~~~~~~~~~~~~~

```

Updating ros2_control target version to match what is currently in [moveit2](https://github.com/ros-planning/moveit2/blob/main/moveit2.repos#L10).

### DIFF
--- a/moveit2_tutorials.repos
+++ b/moveit2_tutorials.repos
@@ -26,4 +26,4 @@ repositories:
   ros2_control:
     type: git
     url: https://github.com/ros-controls/ros2_control.git
-    version: 2.12.1
+    version: 2.13.0


AndyZe (approved):
This makes sense, especially because it matches https://github.com/ros-planning/moveit2/blob/main/moveit2.repos

Thanks!

PR #519, opened 2022-09-26
Default to Cyclone DDS, remove deprecated time-param algorithms

AndyZe:
### Description

All of the tutorials using Move Group Interface are broken a.t.m. due to a FastDDS issue. Recommend a switch to Cyclone DDS as a workaround, as discussed here:  https://github.com/ros-planning/moveit2/issues/1474#issuecomment-1227509627


gavanderhoorn:
I would suggest to clearly state this is a work-around for a known issue, and that this makes all nodes started using this RMW incompatible with any other nodes *not* using Cyclone DDS (due to https://github.com/ros2/rmw_cyclonedds/issues/184).

That last point is not something which is clear to many users, and I would expect novices trying the MoveIt 2 tutorials to be even less aware of something like this.

doc/tutorials/getting_started/getting_started.rst
```diff
@@ -85,6 +85,15 @@ Optional: add the previous command to your ``.bashrc``: ::
    not required and often skipped by advanced users who use more than one
    Colcon workspace at a time, but we recommend it for simplicity.
 
+Switch to Cyclone DDS
+^^^^^^^^^^^^^^^^^^^^^
+
+As of Sep 26, 2022, the default ROS2 DDS has an issue which is fixed by switching to Cyclone:
```
gavanderhoorn, line 91: Pedantic, but the issue isn't really *fixed*, insomuch as it's worked around by avoiding the RMW with the issue completely.

JafarAbdi (approved):

vatanaksoytezer (approved):

PR #526, opened 2022-10-05
Rename MoveItControllerManager

AndyZe:
Description

We had a talk about what is confusing with MoveIt controller managers in July. The only action item that came out of it was renaming MoveItControllerManager->Ros2ControlManager and MoveItMultiControllerManager->Ros2ControlMultiManager.

The reason for this is to distinguish these 2 options from the MoveItSimpleControllerManager, which is a basic action interface and should be used when ros2_control isn't involved.
Merge with...

moveit2 PR: https://github.com/ros-planning/moveit2/pull/1601


AndyZe:
Closing in favor of #531 (I actually forgot I had already started this PR)

PR #533, opened 2022-10-19
Build and host docker images of the tutorials

MikeWrock:
### Description

Beginners to MoveIt that are familiar with Docker may prefer to use a pre-built image rather than set up a workspace, install ROS, MoveIt, and their dependencies as well as completing the steps of the tutorial.

This PR will build and host images on the GHCR such that a user can pull the image and skip all the tutorial instructions other than the ros2 launch commands.

### How to test
The documentation has been updated to include instructions, but they can be summarized as follows:
`wget https://raw.githubusercontent.com/ros-planning/moveit2_tutorials/main/.docker/docker-compose.yml`
OR (since the images are not available on the ros-planning ghcr yet):
`wget https://raw.githubusercontent.com/mikewrock/moveit2_tutorials/PR-prep/.docker/docker-compose.yml`

`DOCKER_IMAGE=rolling-tutorial docker compose run --rm --name moveit2_container gpu`

Test Planning Around Objects:

`ros2 launch moveit2_tutorials demo.launch.py rviz_config:=panda_hello_moveit.rviz & `
`ros2 run hello_moveit hello_moveit`

Test Pick and Place with MoveIt Task Constructor:

`ros2 launch moveit2_tutorials mtc_demo.launch.py &` 
`ros2 launch mtc_tutorial pick_place_demo.launch.py` 

(alternatively you can open a second terminal and use `docker exec -it moveit2_container bash`, `source /opt/ros/rolling/setup.bash`, `source install/setup.bash`, and `ros2 launch mtc_tutorial pick_place_demo.launch.py`)

### Checklist


.docker/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+# Example command:
+#
+#   DOCKER_IMAGE=humble-tutorial docker compose run gpu
```
sea-bass, line 3: I'd include at least the rolling one here, and the cpu option... and maybe a pointer to the doc here too

.docker/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+    . "install/setup.sh" &&\
+    cd "src" &&\
+    ros2 pkg create \
+    --build-type ament_cmake \
+    --dependencies moveit_ros_planning_interface moveit_visual_tools rclcpp \
+    --node-name hello_moveit hello_moveit
+
+######################### Planning Around Objects  #######################################
+
+#Remove the hellow_world tutorial cpp file and replace it with the planning_around_objects file
```
sea-bass, line 33:
```suggestion
#Remove the hello_world tutorial cpp file and replace it with the planning_around_objects file
```

doc/tutorials/quickstart_in_rviz/launch/demo.launch.py
```diff
@@ -1,20 +1,41 @@
...
+from launch_ros.substitutions import FindPackageShare
 from launch.actions import ExecuteProcess
 from ament_index_python.packages import get_package_share_directory
 from moveit_configs_utils import MoveItConfigsBuilder
 
 
 def generate_launch_description():
 
+    declared_arguments = []
+    # UR specific arguments
```
sea-bass, line 16: This comment may be left over from somewhere else, as there is no UR here :)

doc/tutorials/quickstart_in_rviz/launch/demo.launch.py
```diff
@@ -102,16 +125,15 @@ def generate_launch_description():
...
-            static_tf,
-            robot_state_publisher,
-            run_move_group_node,
-            ros2_control_node,
-            joint_state_broadcaster_spawner,
-            arm_controller_spawner,
-            hand_controller_spawner,
-        ]
-    )
+    return nodes_to_start
```
sea-bass, line 139: Would this have to be `return LaunchDescription(nodes_to_start)`?

doc/tutorials/quickstart_in_rviz/launch/panda_hello_moveit.rviz
```diff
@@ -0,0 +1,504 @@
+Panels:
```
sea-bass, line 1: Why do we need a new RViz config here?

Also, it's missing the `MarkerArray` widget subscribing to `/rviz_visual_tools`, if you wanna add that back in.

sea-bass (commented):
This overall looks good!

As someone without a GPU, it look me a little while to find that there was a CPU instruction.

More importantly, I ran into the race condition at startup with the robot's state which is solved by installing and using Cyclone DDS. I would consider using that for the Humble+ images (i.e., everywhere):

```
apt-get install -y ros-rolling-rmw-cyclonedds-cpp
export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
```

If you don't do this, the robot often starts up at zero position and basically all the planning demos here fail.

---

Also, I'm generally wary about recommending people run multiple commands in one terminal with the `&` character as it makes it really hard to kill lingering processes. Would not suggest that in the docs, but rather have people use various terminals. I don't think that is actually typed out anywhere outside the PR description though.

---

Finally, why is it that `docker exec -it moveit2_container bash` requires sourcing stuff? Is this not part of the entrypoint, or does that only happen with `docker run`?

doc/tutorials/quickstart_in_rviz/launch/demo.launch.py
MikeWrock: `nodes_to_start` is the return of the OpaqueFunction, which gets returned from `generate_launch_description` as:

```
    return LaunchDescription(
        declared_arguments + [OpaqueFunction(function=launch_setup)]
    )
```

doc/tutorials/quickstart_in_rviz/launch/panda_hello_moveit.rviz
MikeWrock: I was very hesitant at first to add the second RViz config, especially since it required a rewrite of demo.launch.py (but I think it could be used as a good example of how to pass in strings as launch arguments). The reason I did so was the tutorials instruct the user to add an RvizVisualToolsGui panel and save the changes to overwrite the original file. I wanted to be able to run the launch commands in "Planning Around Objects" without having to include "add an RvizVisualToolsGui panel to RViz" step, so I made a second config file (which I just added the MarkerArray to).

An alternate approach I could take would be to modify the original config file when placing it in the container to include the RvizVisualToolsGui panel and MarkerArray

MikeWrock:
@sea-bass Thanks for the review, really helpful feedback. A few responses here:

> As someone without a GPU, it look me a little while to find that there was a CPU instruction.

I added a bit more to the docker-compose example, but I'm mainly hoping people will see the line in the [docker guide](https://moveit.picknik.ai/main/doc/how_to_guides/how_to_setup_docker_containers_in_ubuntu.html) 

> More importantly, I ran into the race condition at startup with the robot's state which is solved by installing and using Cyclone DDS.

Why is it that I constantly see this being a solution to so many difficult to diagnose problems? What indicators did you see that helped you identify the race condition?

> Also, I'm generally wary about recommending people run multiple commands in one terminal with the & character as it makes it really hard to kill lingering processes

I completely agree and wrote that for ease of use for the reviewers, but its probably best not to mention it and leave that usage to someone who's comfortable with it

> Finally, why is it that docker exec -it moveit2_container bash requires sourcing stuff? Is this not part of the entrypoint, or does that only happen with docker run?

Thank you for asking this, I've always just accepted it as a minor inconvenience but it doesn't need to be. The entrypoint only gets executed during `docker run`, but I added some `source` commands to the bashrc so that entering the container using `-it` will automatically source the `opt` and `install` `setup.bash` as well as set RMW implementation to Cyclone DDS

doc/tutorials/quickstart_in_rviz/launch/panda_hello_moveit.rviz
sea-bass: I think that's fair. Let's keep it :)

sea-bass:
> Why is it that I constantly see this being a solution to so many difficult to diagnose problems? What indicators did you see that helped you identify the race condition?

Fun, isn't it? The indicator was the Panda showing in RViz as starting "upright" in the zero configuration rather than at its home position.

doc/tutorials/quickstart_in_rviz/launch/demo.launch.py
sea-bass: Ah yes, that would do it.

sea-bass (approved):

PR #554, opened 2022-12-02
Delete an unused function

AndyZe:
My compiler warned that this function is unused.


doc/tutorials/quickstart_in_rviz/test/bringup_test.cpp
```diff
@@ -17,11 +17,6 @@ namespace
 __attribute__((no_sanitize("thread")))
 #endif
 #endif
-rclcpp::Node::SharedPtr
-make_node(std::string const& name, rclcpp::NodeOptions const& options)
-{
```
tylerjw, line 22: If you are going to delete this delete the whole annonomous namespace and the ifdef stuff.

PR #515, opened 2022-09-19
Update moveit_cpp tutorial

sjahr:
### Description

Blocked by https://github.com/ros-planning/moveit2/pull/1542

This PR includes small changes to the moveit_cpp tutorial to prevent it from bracking when #1542 gets merged in moveit2

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit2/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #569, opened 2023-01-10
Update moveit_cpp_tutorial because of changes from moveit2#1542 (#515)

jspricke:


rhaschke:
Obviously, https://github.com/ros-planning/moveit2/pull/1542 wasn't backported to Humble. Thus this PR should target `main` instead?

jspricke:
Oh, I see my problem now. Would it be possible to have a rolling alias to main?

PR #571, opened 2023-01-11
Use nightly Debian packages for the tutorial

jspricke:


doc/tutorials/getting_started/getting_started.rst
```diff
@@ -1,40 +1,36 @@
...
-Here we will setup your environment for best running the tutorials. This will create a Colcon workspace, download all of the latest MoveIt source code, and build everything from source to ensure you have the latest fixes and improvements.
-
-Building all the source code of MoveIt can take 20-30 minutes, depending on the CPU speed and available RAM of your computer. If you are on a less performant system, or generally just want to get started quicker, checkout our :doc:`Docker Guide </doc/how_to_guides/how_to_setup_docker_containers_in_ubuntu>`.
+Here we will setup your environment for best running the tutorials. This will create a Colcon workspace and install the latest MoveIt version.
 
 Install ROS 2 and Colcon
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-:ros_documentation:`Install ROS 2 {DISTRO_TITLE}<Installation.html>`.
+^^^^^^^^^^^^^^^^^^^^^^^^
+Follow the :ros_documentation:`Install ROS 2 {DISTRO_TITLE}<Installation.html>` to install ros-humble-ros-base.
```
rhaschke, line 8: Use sphinx variables:
```suggestion
Follow the :ros_documentation:`Install ROS 2 {DISTRO_TITLE}<Installation.html>` to install ros-${DISTRO}-ros-base.
```

doc/tutorials/getting_started/getting_started.rst
```diff
@@ -1,40 +1,36 @@
...
 It is easy to miss steps when going through the ROS 2 installation tutorial. If you run into errors in the next few steps, a good place to start is to go back and make sure you have installed ROS 2 correctly.  One that users commonly forget is to source the ROS 2 install itself.  ::
 
   source /opt/ros/humble/setup.bash
```
rhaschke, line 11:
```suggestion
  source /opt/ros/${DISTRO}/setup.bash
```

doc/tutorials/getting_started/getting_started.rst
```diff
@@ -1,40 +1,36 @@
...
   sudo apt update
   sudo apt dist-upgrade
 
-Install :ros_documentation:`Colcon <Tutorials/Colcon-Tutorial.html#install-colcon>` the ROS 2 build system with `mixin <https://github.com/colcon/colcon-mixin-repository>`_: ::
+Install :ros_documentation:`Colcon <Tutorials/Colcon-Tutorial.html#install-colcon>` the ROS 2 build system: ::
 
   sudo apt install python3-colcon-common-extensions
-  sudo apt install python3-colcon-mixin
-  colcon mixin add default https://raw.githubusercontent.com/colcon/colcon-mixin-repository/master/index.yaml
-  colcon mixin update default
```
rhaschke, line 33: Why do you drop `mixin`? That's a very useful extension.

rhaschke (commented):
Surely binary debian packages will speedup the setup process.
However, this will also enforce nightly ROS packages for the whole system, which might be not desirable for ordinary users...

I suggest to start with released packages.
Only if that should fail, the user should voluntarily choose between a source build or nightly binaries.

jspricke:
Hi Robert, thanks for your review. This is currently more a technology preview proposed by Tyler. The nightly builds are only done for the tutorial packages: https://github.com/moveit/moveit2_packages/blob/main/.github/workflows/build.yaml#L16 and fail for rolling, that's why I added the specific instructions. Dropping mixin is only a proposal to make the tutorial easier to understand, as it is not needed.

rhaschke:
Interesting github action! How do you specify were to deploy to? I couldn't spot any corresponding arguments.

jspricke:
Deployment is currently preconfigured here: https://github.com/jspricke/ros-deb-builder-action/blob/main/action.yaml#L67 I'm planning to make it customizable in future.

tylerjw:
I've wanted to have the tutorials work off the debains packaged through up the upstream binaries. The problem is that very quickly after release, there is always some change people want to reflect in the tutorials. On our client projects, we run into a similar situation and default to building moveit from the source. We generally have a good process to ensure that what is on main in moveit is always usable. This effort should greatly speed up people using our tutorials and others working on client projects.

Another thing is MTC still hasn't been officially released, and we have tutorials here for it, meaning that we always have to build it from source. The GitHub action that @jspricke built handles that nicely too.

tylerjw:
I've tested this locally and got a build error about cyclone dds not being on my system. After following the step at the bottom of the tutorial (manually installing cyclone) it did build. Are we somehow depending on cyclone dds directly in one of our CMake files here but not in our package.xml?

sea-bass (commented):
I started from a fresh ROS 2 Humble Docker image and it worked for me!

No issues switching to CycloneDDS either.

image

However, I don't think we should flat out REPLACE the getting started with these steps, but rather provide alternatives. @MikeWrock already made a Docker container as an alternative in https://github.com/ros-planning/moveit2_tutorials/pull/533, and in that PR he added a separate "how to" and [links from the tutorials](https://github.com/MikeWrock/moveit2_tutorials/blob/main/doc/tutorials/pick_and_place_with_moveit_task_constructor/pick_and_place_with_moveit_task_constructor.rst#:~:text=If%20you%20just%20want%20to%20run%20the%20tutorial%2C%20you%20can%20follow%20the%20%3Adoc%3A%60Docker%20Guide%20%3C/doc/how_to_guides/how_to_setup_docker_containers_in_ubuntu%3E%60%20to%20start%20a%20container%20with%20the%20completed%20tutorial.) for the alternative.

We should probably look to do the same, so we now have clearly communicated local builds with the standard Debian packages, local builds with PickNik nightly Debians, and Docker, for users to choose.

doc/tutorials/getting_started/getting_started.rst
```diff
@@ -49,24 +45,16 @@ Move into your Colcon workspace and pull the MoveIt tutorials source: ::
...
   sudo apt update && rosdep install -r --from-paths . --ignore-src --rosdistro $ROS_DISTRO -y
 
-The next command will configure your Colcon workspace: ::
+The next command will build your Colcon workspace: ::
 
   cd ~/ws_moveit2
-  colcon build --mixin release
-
-This build command will likely take a long time (20+ minutes) depending on your computer speed and amount of RAM available (we recommend 32 GB). If you are short on computer memory or generally your build is struggling to complete on your computer, you can append the argument ``--parallel-workers 1`` to the colcon command above.
+  colcon build
```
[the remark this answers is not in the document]
tylerjw: Maybe the solution would be to provide a "How to use moveit from source" page and link to that and the docker page as alternatives to this?

Having the fastest workflow as the default makes sense. There are advantages to both the from source and docker approaches, but those only apply to some.

doc/tutorials/getting_started/getting_started.rst
```diff
@@ -49,24 +45,16 @@ Move into your Colcon workspace and pull the MoveIt tutorials source: ::
...
-
-  vcs import < moveit2_tutorials/moveit2_tutorials.repos
-
-The import command may ask for your GitHub credentials. You can just press Enter until it moves on (ignore the "Authentication failed" error).
-
 Build your Colcon Workspace
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 The following will install from Debian any package dependencies not already in your workspace. This is the step that will install MoveIt and all of its dependencies: ::
 
   sudo apt update && rosdep install -r --from-paths . --ignore-src --rosdistro $ROS_DISTRO -y
```
eholum, line 52: If you just follow the tutorials blindly this var never gets set. Should make a note to add it to `~/.bashrc` or something, or since we're using humble everywhere else:

```suggestion
  sudo apt update && rosdep install -r --from-paths . --ignore-src --rosdistro humble -y
```

jspricke: you are reading the source, it get's replace in the rendered version.

vatanaksoytezer:
To be honest I am a big fan of Debians and this latest Debian builder. But in the case of tutorials, I am not extremely keen on this change. Tutorials are meant to teach people standard ways of developing and help people get used to the ecosystem. I would prefer to keep standard rosdep / vcs way of building tutorials this instead of debians since people are developing this way everywhere else in ROS / ROS2 ecosystem.

Having said that we should definitely not build moveit2 from source for tutorials. Instead we should create CI jobs for binary builds and promote using rosdep to pull binaries of moveit2 to be used with tutorials. We should at least strictly encourage usage of binaries in released distros.

jspricke:
@vatanaksoytezer I'm completely fine with dropping this but we need to release MTC to switch to official debs.

JafarAbdi:
> @vatanaksoytezer I'm completely fine with dropping this but we need to release MTC to switch to official debs.

I can work on releasing it on Friday or this weekend

PR #572, opened 2023-01-15
Dropped moveit_task_constructor/.repos

rhaschke:
There is no `.repos` file in latest [ros2 branch](https://github.com/ros-planning/moveit_task_constructor/tree/ros2).


JafarAbdi:
I thought we need moveit2 from source for the latest MTC ros2 branch, does it compile with the latest release?

rhaschke:
I don't know. GHA always compiles against the `source` docker image...
However, there is no `.repos` file in the ros2 branch. I don't know since when...

AndyZe (approved):

PR #575, opened 2023-01-17
Bump ros2_control to be compatible with moveit2

jspricke:


v4hn (approved):

rhaschke:
If you want to use a released version, why building it from source at all?
Listing it here, obviously requires regular updating of the release version. Just relying on the latest release would resolve that issue - also in future.

jspricke:
The context is #571 and that the tutorials currently advice to build it from source. If that's no longer needed I would propose to drop the line instead (though even the current version seems to not compile).

PR #578, opened 2023-01-21
Update CI deploy job

tylerjw:
### Description

Reference: https://github.com/ros-planning/moveit_tutorials/commit/4bcd8cca58a56316c7096ee9be92b3b488ffcbf5

@rhaschke would you mind reviewing this change?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,4 +1,4 @@
-name: Build and Deploy Pages
+name: Website
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -10,51 +10,57 @@ on:
...
-      if: ${{ success() }}
-      uses: peaceiris/actions-gh-pages@v3
+
+    - name: Upload pages artifact
+      uses: actions/upload-pages-artifact@v1
+      if: github.repository_owner == 'ros-planning'
       with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
-        publish_dir: ./build/html
-        cname: moveit.picknik.ai
```
henningkayser, line 60: don't we still have to set the cname somewhere?

henningkayser (commented):
Looks correct to me. Is this only a simplification or does it apply any non-obvious changes that we should document?

.github/workflows/deploy.yml
tylerjw: I just pushed a fix for this.

henningkayser (approved):

.github/workflows/deploy.yml
```diff
@@ -10,51 +10,60 @@ on:
...
+        path: build/html
+
+  deploy:
+    if: github.repository_owner == 'ros-planning' && github.ref == 'refs/heads/master'
```
henningkayser, line 60: @tylerjw this should be `refs/heads/main`, correct?

PR #591, opened 2023-02-15
add spacenav example

borongyuan:
### Description

Add a simple spacenav example, only launch file. Need to wait for https://github.com/ros-drivers/joystick_drivers/pull/251 to be merged. There I added an option to make spacenav_node publish TwistStamped directly.

### Checklist

### DIFF
--- a/doc/examples/realtime_servo/launch/servo_spacenav_input.launch.py
+++ b/doc/examples/realtime_servo/launch/servo_spacenav_input.launch.py
@@ -1,0 +1,17 @@
+from launch import LaunchDescription
+from launch_ros.actions import Node
+
+
+def generate_launch_description():
+    return LaunchDescription(
+        [
+            Node(
+                package="spacenav",
+                executable="spacenav_node",
+                name="spacenav",
+                output="screen",
+                parameters=[{"zero_when_static": True, "use_twist_stamped": True}],
+                remappings=[("spacenav/twist", "servo_node/delta_twist_cmds")],
+            ),
+        ]
+    )
--- a/doc/examples/realtime_servo/realtime_servo_tutorial.rst
+++ b/doc/examples/realtime_servo/realtime_servo_tutorial.rst
@@ -31,6 +31,13 @@ Make a service request to start Servo ::
 
 You should be able to control the arm with your controller now, with MoveIt Servo automatically avoiding singularities and collisions.
 
+Using a SpaceMouse
+^^^^^^^^^^^^^^^^^^^^
+
+If you have a 3Dconnexion SpaceMouse, you can send 6DoF Cartesian commands using a single joystick. With the demo still running, in a new terminal, run ::
+
+    ros2 launch moveit2_tutorials servo_spacenav_input.launch.py
+
 Without a Controller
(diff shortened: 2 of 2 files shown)


rhaschke:
Thanks for your effort. According to https://github.com/ros-planning/moveit_tutorials/issues/754#issuecomment-1415648174, `spacenav_node` and the corresponding `spacenavd` are not needed anymore. It would be great if you could try the approach suggested in the linked issue (and thus getting rid of the special spacenav stuff).

AndyZe:
Can you be more specific, @rhaschke? Are you against adding any spacenav functionality completely?

doc/examples/realtime_servo/launch/servo_spacenav_input.launch.py
```diff
@@ -0,0 +1,17 @@
+from launch import LaunchDescription
+from launch_ros.actions import Node
+
+
+def generate_launch_description():
+    return LaunchDescription(
+        [
+            Node(
+                package="spacenav",
```
AndyZe, line 9: An <exec_depend> on `spacenav` should be added to root/package.xml

AndyZe (commented):
Looks good to me. Nice and simple.

rhaschke:
> Can you be more specific, @rhaschke? Are you against adding any spacenav functionality completely?

I suggest to drop spacenav_node specific stuff and operate the space mouse directly via joystick software (only one thing to maintain in future).

borongyuan:
> I suggest to drop spacenav_node specific stuff and operate the space mouse directly via joystick software (only one thing to maintain in future).

Yes, this is what I want. As long as the joystick software can directly publish TwistStamped, other nodes for message conversion can be removed.
Replacing Twist with TwistStamped is useful in future applications. This is also discussed and followed up in other packages.
https://github.com/ros-teleop/teleop_tools/pull/54
https://github.com/ros-planning/navigation2/issues/1594
So it's best to handle this part in joystick_drivers.

gavanderhoorn:
> So it's best to handle this part in joystick_drivers.

not really, I believe.

`joystick_drivers` should care about interfacing with joysticks and gamepads. It should not have anything to do with conversion to arbitrary other messages.

That's what the packages in [ros/ros-teleop](https://github.com/ros-teleop) can be used for.

borongyuan:
> `joystick_drivers` should care about interfacing with joysticks and gamepads. It should not have anything to do with conversion to arbitrary other messages.

The Joy message is a generic message for all types of joysticks. ros-teleop can be used to define the function of any joystick.
While the SpaceMouse is a 6DoF Joystick. So it's reasonable to publish Twist directly. This makes it easier to get started and also reduces delay. We can still use ros-teleop to configure the function of its buttons.
For conversion to other messages, [cartesian_controllers](https://github.com/fzi-forschungszentrum-informatik/cartesian_controllers/tree/master/cartesian_controller_utilities) also provides some references.

AndyZe:
We used teleop_tools in ROS1. It was pretty easy to configure  :+1:

https://github.com/ros-planning/moveit/blob/master/moveit_ros/moveit_servo/config/spacenav_via_teleop_tools.yaml

PR #621, opened 2023-02-21
Set MOVEIT_BRANCH in build_locally.sh

tylerjw:
This sets the moveit branch so the build works and installs doxygen and graphviz.


MarqRazz (approved):
This fixed building locally for me.

Issue #586, opened 2023-02-09
MoveIt Quickstart in RViz does not launch rviz2


zflat:
### Description

When running the demo.launch.py file as instructed at https://moveit.picknik.ai/humble/doc/tutorials/quickstart_in_rviz/quickstart_in_rviz_tutorial.html the following command does not launch rviz 

```
ros2 launch moveit2_tutorials demo.launch.py rviz_tutorial:=true` 
```

However changing `demo.launch.py rviz_tutorial:=true` does launch RViz, for example with
```
ros2 launch moveit2_tutorials demo.launch.py rviz_tutorial:=false
```

This is due to a change made in https://github.com/ros-planning/moveit2_tutorials/commit/e292e4da088420c41269e5301be132a1ca9661ca#diff-a3efd55c9d8991e9a4a31bd3652364beaf91be99fd35307b50f8aa95b077a50c

The fix is to add `rviz_node_tutorial` back to the `nodes_to_start` array in the file `doc/tutorials/quickstart_in_rviz/launch/demo.launch.py`.

### Your environment
* ROS Distro: Humble
* OS Version: Ubuntu 22.04
* Source or Binary build? Source
* If binary, which release version?
* If source, which git commit or tag?
```
~/ws_moveit2/src/moveit2_tutorials$ git log -1
commit e292e4da088420c41269e5301be132a1ca9661ca (grafted, HEAD -> humble, origin/humble)
Author: mergify[bot] <37929162+mergify[bot]@users.noreply.github.com>
Date:   Thu Jan 26 15:10:12 2023 -0500
```

### Steps to reproduce

* Set up as instructed at https://moveit.picknik.ai/humble/doc/tutorials/getting_started/getting_started.html
* Run
  ```
   ros2 launch moveit2_tutorials demo.launch.py rviz_tutorial:=true
   ```
* RViz does not launch

### Expected behaviour
RViz should be launched.

anujay28:
Thanks !!

airbutt:
Took me a day to find that out. You made my day ! by the way binary is not working at all for me, had to build it completely. In Binary there seems some packages missing. But this is another story.....

anujay28:
Yeah , I faced a similar issue. Did you clone the `main` repo  and then build it using `colcon` ?

Albert-lee-7:
I add rviz_node_tutorial to the nodes_to_start array like this
[...]

PR #630, opened 2023-03-03
Update Quickstart in RViz Tutorial Documentation and RViz Config argument name

chancecardona:
### Description

~~Fixes the rviz_tutorial flag in the "MoveIt Quickstart in Rviz" tutorial, as referenced in [Issue 586](https://github.com/ros-planning/moveit2_tutorials/issues/586). Added to declared launch arguments as well, since not setting it on the command line caused an error.~~
Original PR superseded by https://github.com/ros-planning/moveit2_tutorials/pull/635 since problem is mostly fixed in `main`. Edited to include small change to documentation and naming convention going forward; feel free to close if too minor a change.

### Checklist
- [X] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)

### DIFF
--- a/doc/tutorials/quickstart_in_rviz/launch/demo.launch.py
+++ b/doc/tutorials/quickstart_in_rviz/launch/demo.launch.py
@@ -20,6 +20,13 @@ def generate_launch_description():
             description="RViz configuration file",
         )
     )
+    declared_arguments.append(
+        DeclareLaunchArgument(
+            "rviz_tutorial",
+            default_value="False",
+            description="Tutorial flag for empty Rviz config",
+        )
+    )
 
     return LaunchDescription(
         declared_arguments + [OpaqueFunction(function=launch_setup)]
@@ -49,12 +56,30 @@ def launch_setup(context, *args, **kwargs):
         parameters=[moveit_config.to_dict()],
     )
 
-    rviz_base = LaunchConfiguration("rviz_config")
-    rviz_config = PathJoinSubstitution(
-        [FindPackageShare("moveit2_tutorials"), "launch", rviz_base]
+    # RViz
+    tutorial_mode = LaunchConfiguration("rviz_tutorial")
+    rviz_config_file = LaunchConfiguration("rviz_config")
+    rviz_base = os.path.join(get_package_share_directory("moveit2_tutorials"), "launch")
+    rviz_config = PathJoinSubstitution([rviz_base, rviz_config_file])
+    rviz_config_empty = PathJoinSubstitution(
+        [rviz_base, "panda_moveit_config_demo_empty.rviz"]
     )
 
-    # RViz
+    rviz_node_tutorial = Node(
+        package="rviz2",
+        executable="rviz2",
+        name="rviz2",
+        output="log",
+        arguments=["-d", rviz_config_empty],
+        parameters=[
+            moveit_config.robot_description,
+            moveit_config.robot_description_semantic,
+            moveit_config.robot_description_kinematics,
+            moveit_config.planning_pipelines,
+            moveit_config.joint_limits,
+        ],
+        condition=IfCondition(tutorial_mode),
+    )
     rviz_node = Node(
         package="rviz2",
         executable="rviz2",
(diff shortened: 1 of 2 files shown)


doc/tutorials/quickstart_in_rviz/launch/demo.launch.py
```diff
@@ -20,6 +20,13 @@ def generate_launch_description():
             description="RViz configuration file",
         )
     )
+    declared_arguments.append(
+        DeclareLaunchArgument(
+            "rviz_tutorial",
+            default_value="False",
+            description="Tutorial flag for empty Rviz config",
```
sea-bass, line 27: Can you include in the description whether true or false means empty vs. filled out config?

sea-bass (commented):
This works, but what you asked internally about a potentially cleaner way to do this is valid. I was also thinking about that.

See my minor comment, plus whatever falls out from that discussion?

MikeWrock:
@chancecardona , this looks like an issue of the documentation not correctly backporting to humble. On the [main](https://moveit.picknik.ai/main/doc/tutorials/quickstart_in_rviz/quickstart_in_rviz_tutorial.html) (rolling) branch of our docs, the rviz_tutorial argument is dropped. It might be a good idea to point out that there is an empty rviz config should they want one using:

`ros2 launch moveit2_tutorials demo.launch.py rviz_config:=panda_moveit_config_demo_empty.rviz`

I'm more in favour of teaching the user how to specify, modify, save, and select their own rviz config rather than use a `rviz_tutorial` flag for an empty rviz configuration. Which is discussed at the bottom of the tutorial in both the `main` and `humble` version of the docs

chancecardona:
> I'm more in favour of teaching the user how to specify, modify, save, and select their own rviz config rather than use a `rviz_tutorial` flag for an empty rviz configuration, as discussed at the bottom of the tutorial in both the `main` and `humble` version of the docs

Overall I agree with this. I think the `rviz_tutorial` flag is a bit redundant. Editing to update the docs for consistency instead and removing this variable (and the dangling references) unless anyone thinks it should be in `humble`.

zflat:
> @chancecardona , this looks like an issue of the documentation not correctly backporting to humble. On the [main](https://moveit.picknik.ai/main/doc/tutorials/quickstart_in_rviz/quickstart_in_rviz_tutorial.html) (rolling) branch of our docs, the rviz_tutorial argument is dropped. It might be a good idea to point out that there is an empty rviz config should they want one using:
> 
> `ros2 launch moveit2_tutorials demo.launch.py rviz_config:=panda_moveit_config_demo_empty.rviz`
> 
> I'm more in favour of teaching the user how to specify, modify, save, and select their own rviz config rather than use a `rviz_tutorial` flag for an empty rviz configuration, as discussed at the bottom of the tutorial in both the `main` and `humble` version of the docs

I'm also in favor of teaching the user how to run RViz independently of moveit and how to select their own config. I think that launching RViz in the same launch as moveit, controllers, joint state publishers and other manipulation specific nodes is an anti-pattern. Unlike in ROS 1, the order of launching in ROS 2 does not matter (there is no need for the `roscore` process like ROS 1 has). So it is up to the user to decide their preferred workflow including launching RViz once and leaving it running in between launching, stopping, re-building and re-launching manipulation related nodes. Also, moveit and manipulation logic will likely be running on a headless remote machine in a number of systems and launch architecture should reflect that.

doc/tutorials/quickstart_in_rviz/launch/demo.launch.py
chancecardona: Removed this variable as per @MikeWrock's suggestion.

MikeWrock:
@zflat I think your suggestions to discuss how launch Gazebo separately would be a great addition to the tutorials, but for the "MoveIt Quickstart in RViz" tutorial the steps should be as brief and easy as possible

chancecardona:
Edited PR and created a new one with more relevant changes to `humble`. Feel free to close if this is not needed.

sea-bass:
Closing in favor of https://github.com/ros-planning/moveit2_tutorials/pull/635.

PR #521, opened 2022-09-27
Explicitly state which planning pipelines to use in demo launch

stephanie-eng:
### Description
The launch file included with the Quickstart in RViz tutorial uses MoveItConfigsUtils with the panda config files in moveit_resources. By default, MoveItConfigsUtils looks for all `_planning.yaml` files, which picks up lerp_planning.yaml and trajopt_planning.yaml. These files produce the following errors (some logs removed for brevity):

```
[ERROR] [1664267382.867870261] [moveit.ros_planning.planning_pipeline]: Exception while loading planner 'lerp_interface/LERPPlanner': According to the loaded plugin descriptions the class lerp_interface/LERPPlanner with base class type planning_interface::PlannerManager does not exist. Declared types are  chomp_interface/CHOMPPlanner ompl_interface/OMPLPlanner pilz_industrial_motion_planner/CommandPlannerAvailable plugins: chomp_interface/CHOMPPlanner, ompl_interface/OMPLPlanner, pilz_industrial_motion_planner/CommandPlanner
[ERROR] [1664267382.877819227] [moveit.ros_planning_interface.moveit_cpp]: Failed to initialize planning pipeline 'lerp'.
[ERROR] [1664267383.031895256] [moveit.ros_planning.planning_pipeline]: Exception while loading planner 'trajopt_interface/TrajOptPlanner': According to the loaded plugin descriptions the class trajopt_interface/TrajOptPlanner with base class type planning_interface::PlannerManager does not exist. Declared types are  chomp_interface/CHOMPPlanner ompl_interface/OMPLPlanner pilz_industrial_motion_planner/CommandPlannerAvailable plugins: chomp_interface/CHOMPPlanner, ompl_interface/OMPLPlanner, pilz_industrial_motion_planner/CommandPlanner
[ERROR] [1664267383.041862939] [moveit.ros_planning_interface.moveit_cpp]: Failed to initialize planning pipeline 'trajopt'.
```
This PR modifies the launch file to explicitly use only the specified planning pipelines (OMPL, CHOMP, and Pilz), with the intent to prevent the above errors.

PR #634, opened 2023-03-06
Fix launch warnings for MTC tutorial

stephanie-eng:
### Description

This PR fixes the following warnings when launching the MTC demo:
```
[ERROR] [1678136118.485318346] [moveit.ros_planning_interface.moveit_cpp]: Failed to initialize planning pipeline 'trajopt'.
```
```
[ERROR] [1678136118.493854644] [moveit.ros_planning_interface.moveit_cpp]: Failed to initialize planning pipeline 'lerp'.
```
These two errors are caused by MoveIt Configs Utils trying to load all files that match the `*_planning.yaml` files available in MoveIt Resources (see #521). The fix for this is to specify the desired planning pipelines (OMPL, CHOMP, and Pilz), similarly to how it is done in the [Quickstart launch](https://github.com/ros-planning/moveit2_tutorials/blob/main/doc/tutorials/quickstart_in_rviz/launch/demo.launch.py).

```
[rviz2-1] [WARN] [1678136119.251836273] [moveit_ros.robot_model_loader]: No kinematics plugins defined. Fill and load kinematics.yaml!
```
This is caused by not providing RViz with `robot_description_kinematics`, which this PR intends to fix.

### DIFF
--- a/doc/tutorials/pick_and_place_with_moveit_task_constructor/launch/mtc_demo.launch.py
+++ b/doc/tutorials/pick_and_place_with_moveit_task_constructor/launch/mtc_demo.launch.py
@@ -12,6 +12,9 @@ def generate_launch_description():
         MoveItConfigsBuilder("moveit_resources_panda")
         .robot_description(file_path="config/panda.urdf.xacro")
         .trajectory_execution(file_path="config/gripper_moveit_controllers.yaml")
+        .planning_pipelines(
+            pipelines=["ompl", "chomp", "pilz_industrial_motion_planner"]
+        )
         .to_moveit_configs()
     )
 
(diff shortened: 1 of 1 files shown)


stephanie-eng:
As a side note, I believe there are some issues with the tutorial as is, as I'm currently only seeing planning failures from the tutorial. That, however, is an independent issue than this PR (which is just to fix some warnings when running `ros2 launch moveit2_tutorials mtc_demo.launch.py`. Just thought I'd mention it here as a word of warning if anyone tests the tutorial.

sjahr:
> e issues with the tutorial as is, as I'm currently only seeing planning failures from the tutorial. That, however, is an independent issue than this PR

@stephanie-eng Can you open an issue to track this?

sjahr (approved):

PR #644, opened 2023-03-20
Update Dockerfile on the humble branch to use ROS humble

adlarkin:
### Description

The backport in https://github.com/ros-planning/moveit2_tutorials/commit/e292e4da088420c41269e5301be132a1ca9661ca set the `ROS_DISTRO` to `rolling` for the `humble` branch, which would probably cause issues for users who are trying to go through the MoveIt2 tutorials with MoveIt2 humble. This PR sets the `ROS_DISTRO` to `humble` in the Dockerfile so that if users decide to use the Docker setup for the humble tutorials, `ROS_DISTRO` matches the MoveIt2 version.

### Checklist

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,6 +1,6 @@
 # syntax = docker/dockerfile:1.3
 
-ARG ROS_DISTRO=rolling
+ARG ROS_DISTRO=humble
 
 ######################### Tutorial Image  #################################################
 
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ROS_DISTRO: [humble, rolling]
+        ROS_DISTRO: [humble]
     runs-on: ubuntu-latest
     permissions:
       packages: write


MikeWrock (approved):

adlarkin:
@MikeWrock `CI / humble-source` is failing, and I'm not sure why. The error messages in GitHub actions mention [`struct planning_interface::MotionPlanResponse`](https://github.com/ros-planning/moveit2/blob/ff27996375ecdd020e069642368e41f6a94fd005/moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h#L46), but in humble, it looks like the [`PlanningComponent::Plan`](https://github.com/ros-planning/moveit2/blob/b882b0b2c269a192cf72857b84b86a27f7bdfee8/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h#L157-L159) API call returns a [`PlanSolution` struct](https://github.com/ros-planning/moveit2/blob/b882b0b2c269a192cf72857b84b86a27f7bdfee8/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h#L58-L59) instead. In rolling, the `PlanningComponent::Plan` API call does appear to return the `planning_interface::MotionPlanResponse` struct: https://github.com/ros-planning/moveit2/blob/00e10bbebdbef336aacd66fce3f2593844aee834/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/planning_component.h#L199-L201

So, it seems to me like the error stems from rolling code still being used somewhere. I tried modifying https://github.com/ros-planning/moveit2_tutorials/blob/adlarkin/humble_docker_version_fix/.github/workflows/docker-images.yml#L12 to only build humble and not rolling, but I am still seeing the same CI failures. It also looks like humble is used in the industrial_ci for this action (https://github.com/ros-planning/moveit2_tutorials/blob/adlarkin/humble_docker_version_fix/.github/workflows/ci.yaml), so I am not sure what is causing this issue at the moment.

MikeWrock (approved):

adlarkin:
I am going to close this PR - it turns out that for the CI builds that build and push updated tutorial Docker images, the `ROS_DISTRO` is correctly being set via a [build arg](https://github.com/ros-planning/moveit2_tutorials/blob/main/.github/workflows/docker-images.yml#L42-L43) (notice how there are build args for [both rolling and humble](https://github.com/ros-planning/moveit2_tutorials/blob/main/.github/workflows/docker-images.yml#L12), which means that a humble and rolling image will be built and pushed). The real issue here that's causing some of the humble Docker images to actually have rolling installed is upstream (https://github.com/ros-planning/moveit2/pull/2035#issuecomment-1480082496). Once the upstream issue is resolved, we will probably need to re-build the tutorial Docker images since they [depend on the upstream Docker images](https://github.com/ros-planning/moveit2_tutorials/blob/main/.docker/Dockerfile#L7).

I have created an issue so that we can track the progress of getting the tutorial images fixed: #646

PR #653, opened 2023-04-04
distro detected automatically

ozanpali:
### Description

Please explain the changes you made, including a reference to the related issue if applicable
Ros distro is detected automatically.
### Checklist


rhaschke (commented):
`There are many more uses of humble and rolling, which should be replaced with the variable `ROS_DISTRO`. It would be great if we could handle some more of them in this PR`.

tylerjw:
I'm going to close this PR as I just merged a change into the main branch to make it clear that he main branch supports Humble, Iron, and Rolling. I don't think we should change this line because sourcing this file is what sets that environment variable so it wouldn't be set before you source the file.

PR #462, opened 2022-07-08
Adding kinematics cost function tutorial

wyattrees:
### Description

Adding a tutorial that showcases how to use IK cost functions with MoveIt. Please use the branch from [this PR in MoveIt](https://github.com/ros-planning/moveit2/pull/1386)

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit2/pulls) to support the maintainers

PR #669, opened 2023-04-21
Continuation of "Adding kinematics cost function tutorial #462"

sjahr:
### Description

Continues #462 

Original description
```
Adding a tutorial that showcases how to use IK cost functions with MoveIt. Please use the branch from https://github.com/ros-planning/moveit2/pull/1386
```

### Checklist


doc/examples/kinematics_cost_function/src/kinematics_cost_function_tutorial.cpp
```diff
@@ -0,0 +1,256 @@
...
+      max_eef_step, jump_thresh, callback_fn, opts, cost_fn);
+
+  RCLCPP_INFO(LOGGER, "Computed %f percent of cartesian path.", frac.value * 100.0);
+
+  // Once we've computed the points in our Cartesian path, we need to add timestamps to each point for execution.
+  robot_trajectory::RobotTrajectory rt(start_state->getRobotModel(), PLANNING_GROUP);
+  for (const moveit::core::RobotStatePtr& traj_state : traj)
+    rt.addSuffixWayPoint(traj_state, 0.0);
+  trajectory_processing::TimeOptimalTrajectoryGeneration time_param;
+  time_param.computeTimeStamps(rt, 1.0);
```
sjahr, line 235: This causes an error at the moment
```
[kinematics_cost_function_tutorial-1] [INFO] [1682074394.843220509] [kinematics_cost_fn_demo]: Computed 100.000000 percent of cartesian path.
[kinematics_cost_function_tutorial-1] [ERROR] [1682074394.843289210] [moveit_trajectory_processing.time_optimal_trajectory_generation]: No acceleration limit was defined for joint panda_joint1! You have to define acceleration limits in the URDF or joint_limits.yaml
[kinematics_cost_function_tutorial-1] [INFO] [1682074394.843896324] [move_group_interface]: Execute request accepted
[kinematics_cost_function_tutorial-1] [INFO] [1682074394.852646411] [move_group_interface]: Execute request aborted
[kinematics_cost_function_tutorial-1] [ERROR] [1682074394.853204671] [move_group_interface]: MoveGroupInterface::execute() failed or timeout reached
```
Need to figure out why the bounds aren't properly passed down to TOTG through the robot model :thinking:

doc/examples/kinematics_cost_function/launch/kinematics_cost_function_tutorial.launch.py
```diff
@@ -0,0 +1,21 @@
...
+
+    move_group_demo = Node(
+        name="kinematics_cost_fn_tutorial",
+        package="moveit2_tutorials",
+        executable="kinematics_cost_function_tutorial",
+        output="screen",
+        parameters=[
+            moveit_config.robot_description,
+            moveit_config.robot_description_semantic,
+            moveit_config.robot_description_kinematics,
```
sea-bass, line 17: I believe this should fix it

```suggestion
            moveit_config.robot_description_kinematics,
            moveit_config.joint_limits,
```

doc/examples/kinematics_cost_function/src/kinematics_cost_function_tutorial.cpp
```diff
@@ -0,0 +1,256 @@
...
+  target_pose.position.z -= 0.1;
+  target_pose.orientation.w = 0;
+  target_pose.orientation.x = -1.0;
+  Eigen::Isometry3d target;
+  tf2::fromMsg(target_pose, target);
+
+  auto start_state = move_group.getCurrentState(10.0);
+  std::vector<moveit::core::RobotStatePtr> traj;
+  moveit::core::MaxEEFStep max_eef_step(0.01, 0.1);
+  // Here, were effectively disabling the jump threshold for joints. This is not recommended on real hardware.
```
sea-bass, line 219:
```suggestion
  // Here, we're effectively disabling the jump threshold for joints. This is not recommended on real hardware.
```

... but also, the jump threshold for what?

doc/examples/kinematics_cost_function/src/kinematics_cost_function_tutorial.cpp
```diff
@@ -0,0 +1,256 @@
...
+  moveit::core::JumpThreshold jump_thresh(0.0);
+
+  // The trajectory, traj, passed to computeCartesianPath will contain several waypoints toward
+  // the goal pose, target. For each of these waypoints, the IK solver is queried with the given cost function.
+  const auto frac = moveit::core::CartesianInterpolator::computeCartesianPath(
+      start_state.get(), joint_model_group, traj, joint_model_group->getLinkModel("panda_link8"), target, true,
+      max_eef_step, jump_thresh, callback_fn, opts, cost_fn);
```
sea-bass, line 226: Might be worth saying if you don't want to use the cost function (for comparison), how do you leave it out? Would you just omit that last argument?

sea-bass (commented):
Following our latest PR to this repo, since this work is up to date with ROS 2 should it go in the how to guides vs. the examples?

doc/examples/kinematics_cost_function/src/kinematics_cost_function_tutorial.cpp
sjahr: The documentation for this is really good :ok_hand: 
```cpp
/** \brief Struct for containing jump_threshold.

    For the purposes of maintaining API, we support both \e jump_threshold_factor which provides a scaling factor for
    detecting joint space jumps and \e revolute_jump_threshold and \e prismatic_jump_threshold which provide absolute
    thresholds for detecting joint space jumps. */
```

```cpp
/**
     During the computation of the trajectory, it is usually preferred if consecutive joint values do not 'jump' by a
     large amount in joint space, even if the Cartesian distance between the corresponding points is small as expected.
     To account for this, the \e jump_threshold struct is provided, which comprises three fields:
     \e jump_threshold_factor, \e revolute_jump_threshold and \e prismatic_jump_threshold.
     If either \e revolute_jump_threshold or \e prismatic_jump_threshold  are non-zero, we test for absolute jumps.
     If \e jump_threshold_factor is non-zero, we test for relative jumps. Otherwise (all params are zero), jump
     detection is disabled.

     For relative jump detection, the average joint-space distance between consecutive points in the trajectory is
     computed. If any individual joint-space motion delta is larger then this average distance by a factor of
     \e jump_threshold_factor, this step is considered a failure and the returned path is truncated up to just
     before the jump.

     For absolute jump thresholds, if any individual joint-space motion delta is larger then \e revolute_jump_threshold
     for revolute joints or \e prismatic_jump_threshold for prismatic joints then this step is considered a failure and
     the returned path is truncated up to just before the jump.
     **/
```

sea-bass: I understand that users can look at documentation for both these functions I commented on. I'm not asking for you to provide me with the answer to the question, but rather to help newcomers a little bit in the comments so they don't need to go into other files to get basic understanding of the tutorials. For example:

"we're effectively disabling the threshold for joint angle jumps during motion planning"

or

"if you want to compare results without the custom cost function, omit the last argument"

Anyways, these aren't blocking issues by any means. As you said, people can eventually figure this out by reading the source code.

doc/examples/kinematics_cost_function/src/kinematics_cost_function_tutorial.cpp
```diff
@@ -0,0 +1,256 @@
...
+  target_pose.orientation.w = 0;
+  target_pose.orientation.x = -1.0;
+  Eigen::Isometry3d target;
+  tf2::fromMsg(target_pose, target);
+
+  auto start_state = move_group.getCurrentState(10.0);
+  std::vector<moveit::core::RobotStatePtr> traj;
+  moveit::core::MaxEEFStep max_eef_step(0.01, 0.1);
+  // Here, we're effectively disabling the jump threshold for joints. This is not recommended on real hardware.
+  moveit::core::JumpThreshold jump_thresh(0.0);
```
sjahr, line 220: @sea-bass Looks to me like jump checking is disabled completely (see [here](https://github.com/ros-planning/moveit2/blob/dd00f075ffc9b18b6311dc25102d3050cc9d6102/moveit_core/robot_state/src/cartesian_interpolator.cpp#L222)) but I don't know why :shrug: I guess the cost function prevent joint jumps here, since we're minimizing joint motions

sea-bass (approved):

tylerjw (approved):

PR #675, opened 2023-05-05
use latest isaac sim for tutorial

knorth55:
### Description

This PR change to search installed `isaac-sim` and use latest isaac-sim`.

### Checklist

### DIFF
--- a/doc/how_to_guides/isaac_panda/launch/python.sh
+++ b/doc/how_to_guides/isaac_panda/launch/python.sh
@@ -8,7 +8,7 @@
 
 # Script to call the isaac launch system (see python.sh in the ISAAC_SCRIPT_DIR)
 
-ISAAC_SCRIPT_DIR="$HOME/.local/share/ov/pkg/isaac_sim-2022.2.0"
+ISAAC_SCRIPT_DIRS=($(ls -d -- $HOME/.local/share/ov/pkg/isaac_sim-*))
 
 # Prepend the path to all arguments passed in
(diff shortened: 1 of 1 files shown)


JafarAbdi (approved):

doc/how_to_guides/isaac_panda/launch/python.sh
```diff
@@ -18,6 +18,6 @@ do
     NEW_ARGS="${NEW_ARGS} ${CUR_SCRIPT_DIR}/${arg}"
 done
 
-pushd ${ISAAC_SCRIPT_DIR}
+pushd ${ISAAC_SCRIPT_DIRS[${#ISAAC_SCRIPT_DIRS[@]}-1]}
```
MarqRazz, line 21: This does mean that someone will have to manually check that ALL released versions of Isaac, now and into the future, are compatible with this tutorials omnigraph creation script. Pretty sure this is safe right now but Nvidia has been known to make breaking changes to that API. What do you think about adding some checks to verify the version against a known working list?

knorth55: Hmm, your suggestion is right.
I don't trust nvidia either.
I checked this working in `2022.2.1`, so we should check if the package is `2022.2.1` or `2022.2.0`.

MarqRazz: I have also run this back to `1.1` but would have to test it again to make sure it still works. I'm okay with this launch script only supporting `2.0` and `2.1` for now.

knorth55: Thank you for your information. I updated.

PR #681, opened 2023-05-12
Add moveit_msgs to repos file

tylerjw:
### Description

moveit now depends on this change: https://github.com/ros-planning/moveit_msgs/commit/e97a788e80ef2789ded11b3b9b54d0f4b164cc08

cutting a release for moveit_msgs but in the mean time fixing the repos file so it is included


sea-bass (approved):
I recently ran into this, so this for sure is needed.

PR #717, opened 2023-07-21
Join spinner before exit

todoubaba:
### Description

Spinner thread `terminate called without an active exception`.


doc/how_to_guides/pilz_industrial_motion_planner/src/pilz_move_group.cpp
```diff
@@ -179,5 +179,6 @@ int main(int argc, char* argv[])
 
   // Shutdown ROS
   rclcpp::shutdown();
+  spinner.join();
```
sjahr, line 182:
```suggestion
  spinner.join();
  rclcpp::shutdown();
```

sjahr (commented):
Good catch! Just need to call join before shutdown

sjahr (approved):

JafarAbdi (approved):

PR #718, opened 2023-07-21
Remove source builds

sjahr:
### Description

Try install all dependencies from binaries

### Checklist


henningkayser:
as already discussed, dropping source builds is not an option since we want to be able to work on tutorials while we are adding not-yet-released features. I somewhat agree that it would be nice to support the released versions also, but that would come with a lot of syncing issues. I would leave the repos files as they are for now. One thing that we could do is to tag the tutorials every time we cut a new debian release. That way people could run the tagged version with Debians.

PR #720, opened 2023-07-26
Update the displayed tutorial robot

abake48:
### Description

This PR starts the transition to make the displayed robot be the Kinova Gen3 7dof. This coincides with the release of the public ROS 2 driver for the Kinova Gen3 arms.

Due to time constraints, this PR only updates the tutorial section of the docs, excluding the Pick and Place with MoveIt Task Constructor tutorial which will be updated in a follow PR.

To ease the transition from the Panda arm, the Panda tutorial example code has only been renamed, not removed, so users can still reference them if they pull down the repo. 

### Checklist


package.xml
```diff
@@ -60,6 +60,7 @@
   <exec_depend>ros2_control</exec_depend>
   <exec_depend>rviz2</exec_depend>
   <exec_depend>xacro</exec_depend>
+  <exec_depend>kinova_gen3_7dof_robotiq_2f_85_moveit_config</exec_depend>
```
henningkayser, line 63: should we add a config to moveit_resources?

doc/tutorials/your_first_project/your_first_project.rst
```diff
@@ -139,7 +139,7 @@ In place of the comment that says "Next step goes here", add this code:
 
   // Create the MoveIt MoveGroup Interface
   using moveit::planning_interface::MoveGroupInterface;
-  auto move_group_interface = MoveGroupInterface(node, "panda_arm");
+  auto move_group_interface = MoveGroupInterface(node, "manipulator");
```
henningkayser, line 142: I like the more generic naming here

doc/tutorials/your_first_project/panda_hello_moveit.cpp
henningkayser, line 1: do we still need this?

henningkayser (commented):
First superficial pass through GitHub web UI: looks really good so far.
Only more obvious complaint is that all media files have significantly increased in size. Maybe not a huge deal since they are still relatively small.

I'll test this locally before giving the green check mark.

package.xml
abake48: [I tried ](https://github.com/ros-planning/moveit_resources/pull/175)but got push back on this. Didn't want to set the precedence that if we want to modify the tutorial robot, it had to be added to `moveit_resources` which would balloon in size.

doc/tutorials/your_first_project/panda_hello_moveit.cpp
abake48: They aren't being used anymore, but I wasn't sure if we should keep them in the source while we transition to another robot. I could also provide a link within the doc to mention that the tutorial works with another robot

package.xml
henningkayser: I see :/

tylerjw (approved):
I built locally and tried these tutorials, they seem to work for me.

PR #745, opened 2023-08-14
Remove ambiguity in kinematics.rst

uavster:
The current phrasing ("it specifies that a collision check between the two bodies is either not required or wanted") is ambiguous. 

### Description

I guess the intended meaning was "the collision check is not required or the collision is wanted". I removed the "wanted" part to keep the wording consistent with the next sentence and remove the apparent redundancy.

### Checklist

### DIFF
--- a/doc/concepts/kinematics.rst
+++ b/doc/concepts/kinematics.rst
@@ -34,6 +34,6 @@ Allowed Collision Matrix (ACM)
 
 Collision checking is a very expensive operation often accounting for close to 90% of the computational expense during motion planning.
 The ``Allowed Collision Matrix`` or ``ACM`` encodes a binary value corresponding to the need to check for collision between pairs of bodies (which could be on the robot or in the world).
-If the value corresponding to two bodies is set to ``true`` in the ACM, it specifies that a collision check between the two bodies is not required.
+If the value corresponding to two bodies is set to ``true`` in the ACM, no collision check between the two bodies will be performed.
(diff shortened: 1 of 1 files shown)


doc/concepts/kinematics.rst
```diff
@@ -34,6 +34,6 @@ Allowed Collision Matrix (ACM)
 
 Collision checking is a very expensive operation often accounting for close to 90% of the computational expense during motion planning.
 The ``Allowed Collision Matrix`` or ``ACM`` encodes a binary value corresponding to the need to check for collision between pairs of bodies (which could be on the robot or in the world).
-If the value corresponding to two bodies is set to ``true`` in the ACM, it specifies that a collision check between the two bodies is either not required or wanted.
+If the value corresponding to two bodies is set to ``true`` in the ACM, it specifies that a collision check between the two bodies is not required.
```
sjahr, line 37:
```suggestion
If the value corresponding to two bodies is set to ``true`` in the ACM, no collision check between the two bodies will be performed.
```

sjahr (commented):
Good catch! What do you think about being even more specific about that? I think `not required` suggests that it might be an optional check but my understanding is that there won't be a check at all

uavster:
> Good catch! What do you think about being even more specific about that? I think `not required` suggests that it might be an optional check but my understanding is that there won't be a check at all

Much clearer. Thank you!

sjahr (approved):

PR #756, opened 2023-08-23
add pick_ik to moveit2_tutorials.repos

marioprats:
### Description

This change adds pick_ik as a repo needed to build `moveit2_tutorials` from source. Otherwise `rosdep` will pull the precompiled packages for it and its dependencies, including moveit core packages, which may enter in conflict with the source installation.

### Checklist

### DIFF
--- a/moveit2_tutorials.repos
+++ b/moveit2_tutorials.repos
@@ -27,3 +27,7 @@ repositories:
     type: git
     url: https://github.com/ros-planning/moveit_msgs
     version: ros2
+  pick_ik:
+    type: git
+    url: https://github.com/PickNikRobotics/pick_ik
+    version: main


sea-bass (commented):
This is fine, but if we do this we should probably update the source install instructions in the place where we recommend `apt-get remove ros-${ROS_DISTRO}-moveit-*` to similarly remove `pick_ik` binary installs.

marioprats:
I see, but then we should do the same with the other two PickNik-owned repositories that are installed by source: `rosparam_shortcuts` and `rviz_visual_tools`. So we either add the three of them to the `apt-get remove` instructions or neither. I guess until now we have taken the low chance of those being already installed and/or causing trouble, as opposed to MoveIt, which is very likely to be already installed.

Here's another related question: why do we [require MoveIt2 to be installed from source](https://moveit.picknik.ai/humble/doc/tutorials/getting_started/getting_started.html#download-source-code-of-moveit-and-the-tutorials) in order to run the tutorials? Shouldn't the released packages contain all we need?

tylerjw:
Those instructions are there because there was a time when having binary installs would conflict with source builds, that is no longer the case. I and others updated the cmake so that is no longer a problem.

tylerjw (approved):

PR #730, opened 2023-08-01
Remove unused panda files

abake48:
### Description

This PR removes the unused panda files, also provides a fix for CI failures as a result of cache misses

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit2/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #781, opened 2023-10-03
Fix Hello MoveIt copy in Dockerfile

MarqRazz:
This was updated in #739 and then broken again by #730.

Hopefully this will fix CI on `main`.


sea-bass (approved):
Thanks!

PR #782, opened 2023-10-04
Change `Build Tutorial Image` CI job to only build main against Rolling

MarqRazz:
This PR started out trying to fix the CI job `Build Tutorial Image` so that it supported both `humble` and `rolling` but as we investigated further it did not make sense to build both `humble` and `main` when changes were only made to one branch. A follow up PR will be made to the `humble` branch to only build and update the corresponding image.

This is an attempt to fix the `Build Tutorial Image` step for humble. It's currently failing with the following error.

```
#18 517.2 --- stderr: hello_moveit
#18 517.2 /root/ws_moveit/src/hello_moveit/src/hello_moveit.cpp: In function 'int main(int, char**)':
#18 517.2 /root/ws_moveit/src/hello_moveit/src/hello_moveit.cpp:107:36: error: 'std::tuple_element<1, const std::pair<bool, moveit::planning_interface::MoveGroupInterface::Plan> >::type' {aka 'const struct moveit::planning_interface::MoveGroupInterface::Plan'} has no member named 'trajectory'; did you mean 'trajectory_'?
#18 517.2   107 |     draw_trajectory_tool_path(plan.trajectory);
#18 517.2       |                                    ^~~~~~~~~~
#18 517.2       |                                    trajectory_
#18 517.2 gmake[2]: *** [CMakeFiles/hello_moveit.dir/build.make:76: CMakeFiles/hello_moveit.dir/src/hello_moveit.cpp.o] Error 1
```

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -11,11 +11,8 @@ LABEL org.opencontainers.image.description "This container has working versions 
 # Copy MoveIt sources from docker context
 COPY . src/moveit2_tutorials
 
-# Commands are combined in single RUN statement with "apt/lists" folder removal to reduce image size
-# https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers
-RUN --mount=type=cache,target=/root/.ccache/ \
-    # Fetch required upstream sources for building
-    vcs import src < /root/ws_moveit/src/moveit2_tutorials/.github/upstream.repos
+# Fetch required upstream sources for building
+RUN vcs import src < .github/upstream.repos
 
 ######################### Hello World Tutorial  #########################################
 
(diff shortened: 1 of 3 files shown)


MarqRazz:
Thanks for the help @rhaschke!

rhaschke:
You are welcome. The main issue is that the docker/build-push-action uses the current git branch by default:
https://github.com/docker/build-push-action#git-context

However, fixing this via 40dd10e raises new issues... I'm looking into this shortly, but I will handover to you again, if my time window is over.

rhaschke:
Why do you want to build both docker images (for rolling and humble) from the same branch, while they need different source branches:
https://github.com/ros-planning/moveit2_tutorials/blob/6ec65420681b26efb3b2ee57bce79bfddece9922/.github/workflows/docker-images.yml#L12

I suggest building the images only from their corresponding branch... 
If you don't like that approach, I think you will need to sync the Dockerfile from main to humble branch.

I need to continue with other stuff now.

MarqRazz:
> I suggest building the images only from their corresponding branch...

I agree! If you are okay with the change I made I will also open up a PR for the `humble`  branch.

rhaschke (approved):
I reverted some changes that are not needed anymore if we can just use the git context.

PR #783, opened 2023-10-05
Change Build Tutorial Image CI job to only build humble branch against Humble

MarqRazz:
This is the follow up to #782 so that changes to the `humble` branch only build a Humble docker image.

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -8,14 +8,11 @@ FROM moveit/moveit2:${ROS_DISTRO}-source as tutorial_image
 
 LABEL org.opencontainers.image.description "This container has working versions of the tutorials discussed here: https://moveit.picknik.ai/main/doc/tutorials/tutorials.html"
 
-# Copy MoveIt sources from docker context
+# Copy sources from docker context
 COPY . src/moveit2_tutorials
 
-# Commands are combined in single RUN statement with "apt/lists" folder removal to reduce image size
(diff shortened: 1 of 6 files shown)


MarqRazz:
Sorry I assumed we needed to do all of that branch switching to make `humble` work. Where is the context set so that the correct branch is used?

rhaschke:
> I assumed we needed to do all of that branch switching to make `humble` work. Where is the context set so that the correct branch is used?

The docker build action automatically uses the current github context, i.e. the pushed branch / sha:
https://github.com/docker/build-push-action#git-context
> By default, this action uses the [Git context](https://docs.docker.com/engine/reference/commandline/build/#git-repositories), so you don't need to use the [actions/checkout](https://github.com/actions/checkout/) action to check out the repository as this will be done directly by [BuildKit](https://github.com/moby/buildkit).

MarqRazz:
Any ideas to fix the CI failure? Should it block merging this?

I tried to reproduce locally by checking out `moveit/moveit2:humble-release` and building `moveit_visual_tools` on the `ros2` branch and it builds successfully.

rhaschke:
I have no idea, by heart, how to fix this. However, I think this is unrelated to this PR and thus shouldn't block merging.
The main branch also currently fails (due to unrelated reasons):
https://github.com/ros-planning/moveit2_tutorials/actions/runs/6419956713

PR #799, opened 2023-10-25
Remove recommendation for Cyclone due to FastDDS issues

henningkayser:
We don't have any open [issues with FastDDS](https://github.com/ros-planning/moveit2/issues/1874) anymore, everything is running fine per our knowledge. Let's remove this recommendation.


sea-bass (commented):
Should we remove the `export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp` lines from our Docker images and see if CI passes?

MarqRazz:
I agree with @sea-bass but it would probably be best if we always check against both right?

henningkayser:
> I agree with @sea-bass but it would probably be best if we always check against both right?

We should check both, but probably it makes more sense to do this in the moveit2 repo, running our actual tests.
I'll pull the cyclone setup from the tutorials docker image, anyway.

sea-bass (approved):
seems like it works!

sjahr (approved):

PR #813, opened 2023-11-13
Fixed docs: Clone correct moveit2_tutorials branch for Humble

DiegoCarvajal98:
### Description
Added the branch command to clone the humble branch of the `moveit2_tutorials` repo to make sure there are no compatibility issues with ROS 2 Humble.

Please explain the changes you made, including a reference to the related issue if applicable
Added the `-b humble` flag to line

> 50 | git clone https://github.com/ros-planning/moveit2_tutorials

on the Getting started tutorial.

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit2/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #816, opened 2023-11-15
Clone humble branch of tutorials

rhaschke:
People frequently run into issues due to changed API between MoveIt Humble and main, e.g. https://github.com/ros-planning/moveit_task_constructor/pull/508. 

Additionally to more clearly emphasizing, which tutorial to use (unfortunately, the corresponding combobox to select the version was removed at some point), the humble version should only clone the humble branch of the tutorials.


sea-bass:
This was already handled in https://github.com/ros-planning/moveit2_tutorials/pull/813, will merge that and close this as duplicate.

rhaschke:
You should better use the sphinx variable `{DISTRO}` instead of hardcoding those names. This will transfer more easily to other branches...

rhaschke:
Please also consider re-introducing the version-combobox.

Issue #821, opened 2023-11-30
moveit2_tutorials demo.launch.py not working


sjahr:
### Description

Launch the demo.launch.py file fails probably due to an invalid parameter

### Your environment
* ROS Distro: [Rolling]
* OS Version: Ubuntu 22.04
* Source or Binary build? Source
* If source, which git commit or tag? main https://github.com/ros-planning/moveit2_tutorials/commit/71ee840a50794bbfd35e5b63e80fc633d89840ed

### Steps to reproduce
```
ros2 launch moveit2_tutorials demo.launch.py
```

### Expected behaviour
Demo launches

### Backtrace or Console output

```
^^/ws_moveit_tutorials >>> ros2 launch moveit2_tutorials demo.launch.py                                                                                                                                     11:58:58
[INFO] [launch]: All log files can be found below /home/sebastian/.ros/log/2023-11-30-11-58-59-792639-hedgehog12-180049
[INFO] [launch]: Default logging verbosity is set to INFO
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Invalid parameter "isaac_joint_commands"
```

dyackzan:
The main branch of ros2_kortex was recently updated to include some new arguments in the description macros for rendering the kinova and I'm guessing there is now a compatibility issue with the way the kinova is being launched here. I'll investigate, but a short term fix will probably be to check out the commit with hash 740d21b5650a74104fb8c5afb56f7b6f055cdb83 in the ros2_kortex repo (the commit before the kinova macros were updated).

dyackzan:
[...]

PR #826, opened 2023-11-30
Fix CI 

Abishalini:
* Adding `ros2_kortex` as a source build since CI is failing as it does not have the latest changes in `ros2_kortex` repo regarding the planning pipeline refactor.
Fixing the ros2_kortex repo to 
```
[move_group-4] terminate called after throwing an instance of 'rclcpp::ParameterTypeException'
[move_group-4]   what():  expected [string_array] got [string]
```
* Also, added `ros2_robotiq_gripper` repo as a source build to fix the invalid parameter error - https://github.com/ros-planning/moveit2_tutorials/issues/821

* This PR also brings in https://github.com/Kinovarobotics/ros2_kortex/pull/197 to fix the `fake_components` deprecation.

### DIFF
--- a/.github/upstream.repos
+++ b/.github/upstream.repos
@@ -20,8 +20,13 @@ repositories:
     type: git
     url: https://github.com/Kinovarobotics/ros2_kortex.git
     version: main
-  # Remove ros2_robotiq_gripper when rolling binaries are available
+  # Remove ros2_robotiq_gripper when rolling binaries are available.
   ros2_robotiq_gripper:
     type: git
     url: https://github.com/PickNikRobotics/ros2_robotiq_gripper.git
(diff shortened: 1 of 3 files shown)


sjahr (approved):

sea-bass:
@Abishalini I ran into this here: https://github.com/ros2/rcl/issues/1118

I got a suggestion to try out but hadn't got to that. Apparently we may be publishing 2 separate message types to a specific topic and that causes the segfault?

But also, switching to Cyclone DDS gets around this issuem

sea-bass (approved):
The htmlproofer failure is because of the Python bindings. I will look into this in a separate PR to moveit2, so this LGTM.

PR #845, opened 2023-12-18
Check for and fix misspelled MoveIt

stephanie-eng:
We have some misspelled MoveIt occurrences in the tutorials. https://github.com/ros-planning/moveit/pull/2692 has a nice check that tries to catch these. This PR brings in that check and fixes the misspellings.

Maintain the exception for non-words characters in https://github.com/ros-planning/moveit/pull/2692 to prevent failures due to class names like MoveitCpp.

This PR also augments that check for MoveIt 2 so we [respect the space](https://moveit.ros.org/about/press_kit/).

### DIFF
--- a/doc/examples/hybrid_planning/hybrid_planning_tutorial.rst
+++ b/doc/examples/hybrid_planning/hybrid_planning_tutorial.rst
@@ -1,7 +1,7 @@
 Hybrid Planning
 ===============
 
-In this section, you will learn how to use  Moveit 2's *Hybrid Planning* feature.
+In this section, you will learn how to use  MoveIt 2's *Hybrid Planning* feature.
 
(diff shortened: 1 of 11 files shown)


doc/examples/hybrid_planning/hybrid_planning_tutorial.rst
```diff
@@ -1,7 +1,7 @@
 Hybrid Planning
 ===============
 
-In this section, you will learn how to use  Moveit 2's *Hybrid Planning* feature.
+In this section, you will learn how to use  MoveIt 2's *Hybrid Planning* feature.
```
sea-bass, line 4:
```suggestion
In this section, you will learn how to use MoveIt 2's *Hybrid Planning* feature.

sea-bass (approved):
Cool automated check!

PR #849, opened 2023-12-20
Build collision scene example

stephanie-eng:
### Description

Currently, the collision_environments subdirectory (which contains the collision scene example) isn't being built. The collision scene example is used in the [CHOMP tutorial](https://moveit.picknik.ai/main/doc/how_to_guides/chomp_planner/chomp_planner_tutorial.html), which asks the user to run: `ros2 run moveit2_tutorials collision_scene_example`.

Current behaviour:
```bash
$ ros2 run moveit2_tutorials collision_scene_example         
No executable found
```

Expected behaviour after fix:
```bash
$ ros2 run moveit2_tutorials collision_scene_example         
```
The above command should not produce the `No executable found` output. Instead, you should see some boxes in the collision environment. Here's what it looks like when running through the tutorial:
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,7 @@ endforeach()
 include_directories(${THIS_PACKAGE_INCLUDE_DIRS})
 
 # add_subdirectory(doc/examples/bullet_collision_checker)
-# add_subdirectory(doc/examples/collision_environments)
+add_subdirectory(doc/examples/collision_environments)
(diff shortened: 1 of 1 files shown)


MikeWrock (approved):
I tested this change in the tutorial docker container and it worked exactly as described. Thanks @stephanie-eng

sea-bass (approved):

PR #852, opened 2024-01-11
Upgrade the version of ruby used in CI

tylerjw:
### Description

CI seems to be failing because we are using an old version of ruby that our dependencies for htmlprofer no longer support. Locally, on Ubuntu 22.04, the version of ruby I have that works is 3 so I am trying to set the version in CI to 3 to see if it fixes it.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -32,7 +32,7 @@ jobs:
 
     - uses: ruby/setup-ruby@v1
       with:
-        ruby-version: '2.7'
+        ruby-version: '3'
 
     - name: Run htmlproofer.sh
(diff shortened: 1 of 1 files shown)


sea-bass:
Re: rolling source failure -- I thought I bloomed `pick_ik` to rolling, is it just not yet available?

sea-bass (approved):
Anyways, this PR fixes the issue it set out to.

tylerjw:
> Re: rolling source failure -- I thought I bloomed `pick_ik` to rolling, is it just not yet available?

Rolling distribution from the buildfarm is currently broken. Working on fixing that separately.

PR #854, opened 2024-01-15
Update sphinx version for CI to pass on main

Abishalini:
Currently, the `htmlproofer` and `upload_site_artifacts` CI jobs are failing on main due to 
```
Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 pip
-docutils==0.18.0
+docutils==0.18.1
 sphinx>=5.0.0
 sphinx-tabs==3.4.4
 sphinx-multiversion
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 pip
 docutils==0.16
 sphinx>=5.0.0
-sphinx-tabs
+sphinx-tabs==3.4.4
 sphinx-multiversion
 sphinx-rtd-theme
 sphinx-copybutton
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 pip
-docutils==0.16
+docutils==0.18.0
 sphinx>=5.0.0
 sphinx-tabs==3.4.4
 sphinx-multiversion
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 pip
 docutils==0.16
-sphinx
+sphinx>=5.0.0
 sphinx-tabs
 sphinx-multiversion
 sphinx-rtd-theme
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 pip
 docutils==0.16
-sphinx==4.3.2
+sphinx
 sphinx-tabs
 sphinx-multiversion
 sphinx-rtd-theme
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
 pip
 docutils==0.16
-sphinx==4.3.2
+sphinx
```
sea-bass, line 3: To minimize breakages, do you maybe want to specify a version after all?

The one that was downloaded was 7.32 according to https://github.com/ros-planning/moveit2_tutorials/actions/runs/7531539359/job/20500319203?pr=854#step:4:82

You can also use something like `sphinx>=5.0.0` if you just want to skirt around the immediate error but give users wiggle room when building locally.

Generally for these things it's good for the pip requirements to be pinned to actual versions and for any updates to be deliberate.

Abishalini:
In the `htmlproofer` job, it is running Sphinx v5.3.0. The error now is
```
Warning, treated as error:
Invalid configuration value found: 'language = None'. Update your configuration to a valid language code. Falling back to 'en' (English).
```
In the `upload_site_artifacts` job, it is running Sphinx v4.3.2. Not sure what is going on here

Abishalini:
How is it decided which version of packages to use?

If v7.2 is downloaded, why does it use v5.3.0?
```
Collecting sphinx>=5.0.0 (from -r requirements.txt (line 3))
  Downloading sphinx-7.2.6-py3-none-any.whl.metadata (5.9 kB)
```

The lastest error in `htmlproofer` job is 
```
Exception occurred:
  File "/usr/local/lib/python3.10/dist-packages/sphinx_tabs/tabs.py", line 341, in setup
    app.add_javascript(path)
AttributeError: 'Sphinx' object has no attribute 'add_javascript'
```
The attribute `add_javscript` is deprecated and the latest version of `sphinx-tabs` does not use it.

It looks like the latest version of `sphinx-tabs` is downloaded but not used?
```
Collecting sphinx-tabs (from -r requirements.txt (line 4))
  Downloading sphinx_tabs-3.4.4-py3-none-any.whl.metadata (6.3 kB)
```

sea-bass:
Ah, seems like the pinned docutils version is causing harm lol https://github.com/ros-planning/moveit2_tutorials/actions/runs/7532274156/job/20502539940?pr=854#step:4:102

requirements.txt
```diff
@@ -1,7 +1,7 @@
 pip
 docutils==0.16
-sphinx==4.3.2
-sphinx-tabs
+sphinx>=5.0.0
+sphinx-tabs==3.4.4
```
sea-bass, line 4:
```suggestion
docutils
sphinx==5.3.0
sphinx-tabs==3.4.4
```

requirements.txt
```diff
@@ -1,7 +1,7 @@
 pip
-docutils==0.16
-sphinx==4.3.2
-sphinx-tabs
+docutils==0.18.0
```
sea-bass, line 2: lol https://github.com/sphinx-doc/sphinx/issues/10596

```suggestion
docutils==0.18.1
```

sea-bass (approved):
We'll need to rebuild the Docker images to get these other things passing, so just go for it.

PR #855, opened 2024-01-15
Update requirements.txt for humble branch

Abishalini:
Need this update to get CI passing for https://github.com/ros-planning/moveit2_tutorials/pull/854

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 pip
-docutils==0.16
+docutils==0.18.1
 sphinx>=5.0.0
 sphinx-tabs==3.4.4
 sphinx-multiversion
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 pip
 docutils==0.16
```
sea-bass:
```suggestion
docutils==0.18.1
```

sea-bass (commented):
Don't we also want to add the language in the `conf.py` here as well?

sea-bass (approved):

Abishalini:
Force-merging since the failing CI jobs are cloning the humble branch instead of using the current branch.

PR #859, opened 2024-01-16
Ignore MoveIt and MoveIt Core Packages from docs Imports

eholum:
### Description

I think @Abishalini was on the right track with https://github.com/ros-planning/moveit2_tutorials/pull/857, we may just need to explicitly exclude the core package under moveit.

### Checklist


sea-bass (approved):
SEND IT

PR #904, opened 2024-05-03
Change max distance in Connect stages in MTC tutorial

sea-bass:
### Description

I was just trying to get MoveIt 2 tutorials working on Ubuntu 24.04 / Rolling and came back to this long-overdue fix.

This PR implements the suggestion from @rhaschke in https://github.com/moveit/moveit_task_constructor/issues/540#issuecomment-1988467097, which indeed fixes things on my end as well.


rhaschke:
I think, it would be more useful to change the default setting in the Connect stage as I did in https://github.com/moveit/moveit_task_constructor/pull/542. Unfortunately, nobody approved this PR yet.
By the way, I noticed, that Humble doesn't need this relaxation of the threshold. Hence, there was some change in the main branch of MoveIt2, which causes this issue. The best solution would be to revert this change in MoveIt2!
Closing here.

PR #916, opened 2024-06-24
Use venv in CI

stephanie-eng:
### Description

Fix this error:

```
 error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.
```

This should make CI pass the "Install Python Dependencies" step.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,9 +22,12 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Install Python dependencies
+      shell: bash
       run: |
         sudo apt-get update -y
-        sudo apt-get install -y python3-pip
+        sudo apt-get install -y python3-pip python3-venv
+        python3 -m venv .venv
+        source .venv/bin/activate
         pip3 install --upgrade --requirement requirements.txt
 
     - name: Install doxygen and graphviz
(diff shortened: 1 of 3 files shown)


sea-bass (approved):
You may need to add a note to the `htmlproofer.sh` instructions that this needs to be run in a virtual environment. Otherwise, I think we should go for it and fix the actual build errors separately.

stephanie-eng:
I updated a doc page and added in a comment. It looks like that doc page I updated has other issues (the htmlproofer script requires ruby to be installed, the requirements currently say Galactic), but those changes can be done in separate PRs - those issues are not particularly related to the virtual environment stuff.

sea-bass (approved):

PR #920, opened 2024-07-08
Fix htmlproofer CI job on Ubuntu 24.04

sea-bass:
### Description

In Ubuntu 24.04, you can't `pip install` into System Python, encouraging you to use a virtual environment.

As a stop-gap, we can set an environment variable to allow this, which I think is fine just for the docs build job.

### Checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -62,6 +62,7 @@ jobs:
       run: |
         sudo apt-get update -y
         sudo apt-get install -y python3-pip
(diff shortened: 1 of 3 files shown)


sea-bass:
OK well now the job runs correctly, but there are weird failures

stephanie-eng:
Is this meant as an alternative to #916 ?

sea-bass:
> Is this meant as an alternative to #916 ?

I did not even see that one! I think yours is more principled, so I'm happy to close this one... woops.

PR #913, opened 2024-06-07
Temporarily depend on PickNik fork to fix missing gazebo_ros2_control rosdep

henningkayser:
See https://github.com/moveit/moveit2/issues/2864

I also triggered a docker image build on this branch.

PR #921, opened 2024-07-08
Use fork of `ros2_kortex` that does not depend on Gazebo classic

sea-bass:
### Description

In Ubuntu 24.04, Gazebo classic is no longer supported or available from binaries. Due to the `ros2_kortex` repo still using this as a dependency, CI has been failing.

See: https://github.com/Kinovarobotics/ros2_kortex/issues/217 and https://github.com/Kinovarobotics/ros2_kortex/pull/228

So, to have this work, we (and all consumers of this package on Jazzy onwards) have to get around this.

### Checklist

### DIFF
--- a/.github/upstream.repos
+++ b/.github/upstream.repos
@@ -16,10 +16,12 @@ repositories:
     url: https://github.com/PickNikRobotics/rviz_visual_tools.git
     version: ros2
   # Remove ros2_kortex when rolling binaries are available.
+  # Need to use this fork since ros2_kortex depends on Gazebo classic packages:
+  # https://github.com/Kinovarobotics/ros2_kortex/issues/217
   ros2_kortex:
     type: git
-    url: https://github.com/Kinovarobotics/ros2_kortex.git
-    version: main
+    url: https://github.com/sea-bass/ros2_kortex.git
+    version: remove-gazebo-classic
   # Remove ros2_robotiq_gripper when rolling binaries are available.
   ros2_robotiq_gripper:
     type: git
@@ -30,9 +32,3 @@ repositories:
     type: git
     url: https://github.com/tylerjw/serial.git
     version: ros2
-  # This needs to be included due to the ros2_kortex repo:
-  # https://github.com/Kinovarobotics/ros2_kortex/issues/217
-  gazebo_ros2_control:
-    type: git
-    url: https://github.com/ros-controls/gazebo_ros2_control
-    version: master
--- a/.github/upstream.repos
+++ b/.github/upstream.repos
@@ -32,3 +32,8 @@ repositories:
     type: git
     url: https://github.com/tylerjw/serial.git
     version: ros2
+  # Doesn't seem to be available from binaries yet  
+  gz_ros2_control:
+    type: git
+    url: https://github.com/ros-controls/gz_ros2_control.git
+    version: master
--- a/.github/upstream.repos
+++ b/.github/upstream.repos
@@ -32,7 +32,7 @@ repositories:
     type: git
     url: https://github.com/tylerjw/serial.git
     version: ros2
-  # Doesn't seem to be available from binaries yet  
+  # Doesn't seem to be available from binaries yet
   gz_ros2_control:
     type: git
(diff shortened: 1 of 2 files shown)


sea-bass:
@henningkayser where do these dependencies come from? I added to the `.repos` file and it still is failing to be picked up. Does it come from the `moveit2` repo? If so, it's weird to have to add the dependencies there...

sea-bass:
Ugh, no, this then has even more transitive dependencies on `gazebo_ros` and `gazebo_dev` which are not available on Ubuntu 24.04. This will not be a viable option, and instead it's preferable to fix the `ros2_kortex` repo.

I tried doing that, but the maintainers seem reluctant: https://github.com/Kinovarobotics/ros2_kortex/pull/228

sea-bass:
This is now failing because the version of `ros2_control` from binaries doesn't match the latest `gz_ros2_control` ... I give up for now, too many breakages.

stephanie-eng (commented):
For what it's worth, I tried adding ros2_control as a source-built dependency on top of your branch, but it looks like robotiq_driver is not compatible with the latest ros2_control:
https://github.com/stephanie-eng/moveit2_tutorials/actions/runs/9884753382/job/27301653148

I'm also waving the white flag on this for now, so I figured I'd share my findings to whomever tries this again.

henningkayser:
Thank you for your effort! I started a similar thing already here #913. Unfortunately, there is currently just not enough capacity to properly tackle the tutorial issues. I'll still pin this as high priority until we can assign someone to get this fixed.

sea-bass:
> For what it's worth, I tried adding ros2_control as a source-built dependency on top of your branch, but it looks like robotiq_driver is not compatible with the latest ros2_control: https://github.com/stephanie-eng/moveit2_tutorials/actions/runs/9884753382/job/27301653148
> 
> I'm also waving the white flag on this for now, so I figured I'd share my findings to whomever tries this again.

I actually also fixed the robotiq package...

https://github.com/PickNikRobotics/ros2_robotiq_gripper/pull/54

We basically need that landed, a new ros2_control sync or source build, and the same fix for moveit_resources:

https://github.com/moveit/moveit_resources/pull/200

Believe me, I tried...

stephanie-eng:
Well, the rolling sync is done, so we shouldn't need a ros2_control source build... That's one thing out of the way.

stephanie-eng:
We failed again due to an invalid parameter (`sim_ignition`). I think it's due to a stray `sim_ignition` passed into the gripper xacro ([this guy](https://github.com/sea-bass/ros2_kortex/blob/remove-gazebo-classic/kortex_description/grippers/robotiq_2f_85/urdf/robotiq_2f_85_macro.xacro#L30C9-L30C24)), where you have since been renamed it to `sim_gazebo` ([here](https://github.com/PickNikRobotics/ros2_robotiq_gripper/blob/main/robotiq_description/urdf/robotiq_2f_85_macro.urdf.xacro#L8)).

sea-bass:
> We failed again due to an invalid parameter (`sim_ignition`). I think it's due to a stray `sim_ignition` in ros2_kortex passed into the gripper xacro ([this guy](https://github.com/sea-bass/ros2_kortex/blob/remove-gazebo-classic/kortex_description/grippers/robotiq_2f_85/urdf/robotiq_2f_85_macro.xacro#L30C9-L30C24)), where you have since been renamed it to `sim_gazebo` ([here](https://github.com/PickNikRobotics/ros2_robotiq_gripper/blob/main/robotiq_description/urdf/robotiq_2f_85_macro.urdf.xacro#L8)).
> 
> @sea-bass it's your fork, so you get to be the arbiter if the above is correct or not.

You got it!

When I first put up the `ros2_kortex` PR I told myself I wouldn't mess with any other repos in the process, but then I found I had to for the numerous reasons in this thread... so I ended up changing the Robotiq repo after all!

sea-bass:
So the next issue is that the old bringup test is trying to launch RViz that that node is crashing on the CI runner... is this a new thing?

sea-bass:
IT PASSED!!!!!!!!!!

stephanie-eng (approved):

PR #933, opened 2024-07-31
Fix docs build

sea-bass:
### Description

This PR makes an attempt at "fixing" the docs build.

I had to modify the `doxylink` repo (see https://github.com/sphinx-contrib/doxylink/pull/59) since there was an exception without this, and also removed/replaced some old links to make htmlproofer happy.

While the exception was fixed with my PR above, there are still some more build warnings, so I had to remove the `-W` flag from the `sphinx-build` call in `htmlproofer.sh`.

I figured it would be better if the docs actually built and then someone else can try fix the warnings (or change the whole docs build system, which is probably due at this point)

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ build
 Gemfile
 native_build
 *.gdb
+.venv
 # For local rst builds
 _build/
--- a/doc/examples/hybrid_planning/hybrid_planning_tutorial.rst
+++ b/doc/examples/hybrid_planning/hybrid_planning_tutorial.rst
@@ -36,7 +36,7 @@ To start the hybrid planning demo simply run: ::
 
 You should see a similar behavior as in the example GIF above without the replanning.
 
-To interact with the architecture you simply need to send a *Hybrid Planning Request* to an action server offered by the *Hybrid Planning Manager* as seen in the :moveit_codedir:`hybrid_planning_test_node <moveit_ros/hybrid_planning/test/hybrid_planning_demo_node.cpp#L245>`.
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -7,5 +7,6 @@ build
 Gemfile
 native_build
 *.gdb
+.venv
```
sea-bass, line 10: @stephanie-eng adding to gitignore since following the new instructions will create this folder locally

doc/examples/hybrid_planning/hybrid_planning_tutorial.rst
```diff
@@ -36,7 +36,7 @@ To start the hybrid planning demo simply run: ::
 
 You should see a similar behavior as in the example GIF above without the replanning.
 
-To interact with the architecture you simply need to send a *Hybrid Planning Request* to an action server offered by the *Hybrid Planning Manager* as seen in the :moveit_codedir:`hybrid_planning_test_node <moveit_ros/hybrid_planning/test/hybrid_planning_demo_node.cpp#L245>`.
+To interact with the architecture you simply need to send a *Hybrid Planning Request* to an action server offered by the *Hybrid Planning Manager*.
```
sea-bass, line 39: @sjahr This file was removed from the `moveit2` repo last year with the promise that it would eventually be moved to this repo with a refresh of the Hybrid Planning tutorials. I assume this whole file would then get revisited.

.gitignore
stephanie-eng: Good point, thanks for the fix

stephanie-eng (approved):
I can now build the tutorials (well, minus the fact that the build_locally script is broken)! Thank you!

henningkayser (approved):

PR #937, opened 2024-08-01
Install Gazebo as part of Tutorials image build

sea-bass:
### Description

There have been issues building the tutorials image because now some of the packages (namely Robotiq and ros2_kortex) depend on Gazebo being installed.

This will work for Jazzy/Rolling, but I think there need to be args introduced to make this work on other distros (or even changing `gz` for `ignition`)

So you will then need to pass in this `GZ_VERSION` arg in the `moveit2` CI job here: https://github.com/moveit/moveit2/blob/1d93035fc4fbc2831916bce26d7b03e15b8a86d5/.github/workflows/tutorial_docker.yaml#L19

NOTE: You should see the effects of this in the post-merge action for this PR. I've also dispatched the workflow so you can see it on this branch: https://github.com/moveit/moveit2_tutorials/actions/runs/10207096230

(all previous failing runs can be found here: https://github.com/moveit/moveit2_tutorials/actions/workflows/docker.yml)

### Checklist

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,12 +1,15 @@
 # syntax = docker/dockerfile:1.3
 
 ARG ROS_DISTRO=rolling
+ARG GZ_VERSION=8
 
 ######################### Tutorial Image  #################################################
 
 FROM moveit/moveit2:${ROS_DISTRO}-source as tutorial_image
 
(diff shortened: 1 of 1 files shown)


sea-bass:
An alternative to all this, btw, is to remove the dependency on `ros_gz_sim` from the `ros2_kortex` package... while that wouldn't be "proper", as it factually depends on Gazebo, you could consider it an optional dependency given it's just an option in some of the xacros.

sjahr (approved):
Thanks for fixing this. I think for now this is a good way forward

PR #948, opened 2024-08-25
Fix image link in documentation

KarthiAru:
This commit removes a stray image link (https://moveit.ros.org/assets/images/roadmap.png) that was found in the documentation. The link was not relevant to the content.


sea-bass (changes requested):
We just have to change the `https://` link to `http://` and the link will be good.

KarthiAru:
@sea-bass Thanks. I've updated the PR.

sea-bass (approved):

PR #950, opened 2024-08-25
Added code to update submodules during installation

KarthiAru:
Added code to automatically initialize and update all submodules recursively. This change addresses build errors related to un-initialized submodules, specifically issues with the `pybind11` submodule. The script checks each directory for the presence of `.gitmodules` and, if found, performs the necessary submodule updates. This ensures all dependencies are correctly set up before build, improving the reliability of our build process and reducing setup errors for new users.

### DIFF
--- a/doc/tutorials/getting_started/getting_started.rst
+++ b/doc/tutorials/getting_started/getting_started.rst
@@ -66,23 +66,10 @@ Move into your Colcon workspace and pull the MoveIt tutorials source, where ``<b
 
 Next we will download the source code for the rest of MoveIt: ::
 
-  vcs import < moveit2_tutorials/moveit2_tutorials.repos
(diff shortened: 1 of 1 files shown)


sea-bass (changes requested):
As this is specific to only setting up submodules for the MoveIt Task Constructor python bindings, which are optional, I don't think this should take up more space in the default tutorials instructions.

But also, there is a simpler solution to all this, which is to add the `--recursive` flag to `vcs import`:

```suggestion
  vcs import --recursive < moveit2_tutorials/moveit2_tutorials.repos
```

^  maybe this is okay to update?

KarthiAru:
@sea-bass Thanks. I didn't know this simpler option existed. I've updated the PR.

sea-bass (approved):

PR #959, opened 2024-09-06
Getting Started - tutorial changes.rst

alejandrojginerd:
I noticed I was able to build the project (getting started tutorial) without ccache installed. Therefore, every build took over 20 mins.

### Description

Added a note in the tutorial for people to install ccache and reduce subsequent builds.

### DIFF
--- a/doc/tutorials/getting_started/getting_started.rst
+++ b/doc/tutorials/getting_started/getting_started.rst
@@ -95,6 +95,17 @@ This build command will likely take a long time (20+ minutes) depending on your 
   If you are low on computer memory, or if the build is generally having trouble completing on your computer,
   you can try appending ``--executor sequential`` to the ``colcon`` command above to build only one package at a time, or ``--parallel-workers <X>`` to limit the number of simultaneous builds. For even more limited machines, you can try running ``MAKEFLAGS="-j4 -l1" colcon build --executor sequential``.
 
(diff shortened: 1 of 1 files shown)


doc/tutorials/getting_started/getting_started.rst
```diff
@@ -95,6 +95,17 @@ This build command will likely take a long time (20+ minutes) depending on your
   If you are low on computer memory, or if the build is generally having trouble completing on your computer,
   you can try appending ``--executor sequential`` to the ``colcon`` command above to build only one package at a time, or ``--parallel-workers <X>`` to limit the number of simultaneous builds. For even more limited machines, you can try running ``MAKEFLAGS="-j4 -l1" colcon build --executor sequential``.
 
+.. note::
+
+   To greatly reduce the build time of subsecuent builds, make sure you have installed and properly configured ``ccache`` on your compiler.
```
sea-bass, line 100:
```suggestion
   To greatly reduce the build time of subsequent builds, you can optionally set up ``ccache`` on your system.

**Review by sea-bass — COMMENTED** [review]
Thank you for this tip! While this works, I would bias towards using the more "official" colcon mixins workflow that is already documented earlier in this file.

You could for example do this instead:

```bash
colcon build --mixin release ccache
```

Issue #994, opened 2024-12-18
Update all headers to `.hpp`


sea-bass:
After https://github.com/moveit/moveit2/pull/3113, this repo should update all the MoveIt `.h` headers to `.hpp`.

Great chance for someone to help out here!

CihatAltiparmak:
Hi, i want to help you out. For this, we should minimize warnings which is due to other non-updated repositories like moveit_task_constructor and moveit_visual_tools. I've also send PR of these repositories to remove obsolete headers. Additionaly, we should also think about not only updating headers but also updating documentation.

CihatAltiparmak:
Let's close bc looks like it's successfully sorted out.

PR #997, opened 2024-12-20
Enhancement/use hpp for headers based on PR https://github.com/moveit/moveit2/pull/3113

CihatAltiparmak:
### Description

Addresses #994 

### Checklist


mikeferguson:
Looks like CI is failing because @sea-bass maybe did a force push to his branch of ros2_kortex?

mikeferguson:
Also - we aren't going to want to merge this until we get the headers actually released onto the binaries (humble is out, Jazzy sync looks to be happening next monday? not sure about rolling)

CihatAltiparmak:
I agree with you because, as i mentioned, i have to fix documentation so that documentation can show the best practices to new users.

sea-bass:
> Looks like CI is failing because @sea-bass maybe did a force push to his branch of ros2_kortex?

I definitely did do that. Many times.

sea-bass:
> > What's the best way to resolve this?
> 
> I think we need to rebuild the tutorial docker image - they automatically rebuild on Sunday morning I think, I'm not sure how to trigger it manually...

Ah yes, this is done through the moveit repo actions. I'm on it.

sea-bass (approved):
Given both Jazzy and Rolling branches of MoveIt 2 have .hpp files, I am OK approving this as is.

mikeferguson:
> Given both Jazzy and Rolling branches of MoveIt 2 have .hpp files, I am OK approving this as is.

Just a note that Jazzy debs aren't out yet with those headers (ETA monday according to ROS discourse)

sea-bass:
Yeah we can wait.

I didn't release Rolling by the way... so those binaries won't work next sync unless we do.

mikeferguson:
> I didn't release Rolling by the way... so those binaries won't work next sync unless we do.

Rolling has 2.12.0 in testing, but not yet sync'd (we just didn't put 2.12.1 in, which includes the fix for checkstartstate)

sea-bass:
Right. I should know better. Never mind!

CihatAltiparmak:
Hi @sea-bass , I have also updated documentation manually. I think we should wait until merging corresponding PR in moveit_visual_tools. Thus we can early detect potential issues in this PR. The documentation doesn't consist of source codes.

sea-bass:
Sounds good. If you can tag me in the other related PRs when ready for review, I'll take a look.

But for sure let's merge after the Jazzy sync.

Also, for those other repos that may not have distinct branches between humble and later, there are conditional includes we can add -- like was done in moveit2.

sea-bass:
Jazzy sync is in -- shall we merge this?

CihatAltiparmak:
I've updated my branch, this will build package without cache. Let's see if there is any warning in CI. And then good to roll ahead.

PR #999, opened 2024-12-31
Remove pick_ik as a dependency

sea-bass:
### Description

If people want to use pick_ik, it can still be installed optionally. But it's a bit wasteful to have it be built all the time.

### Checklist

### DIFF
--- a/doc/how_to_guides/pick_ik/pick_ik_tutorial.rst
+++ b/doc/how_to_guides/pick_ik/pick_ik_tutorial.rst
@@ -78,7 +78,7 @@ robot's config directory to use pick_ik as the IK solver. ::
 
 .. note::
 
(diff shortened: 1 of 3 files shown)


JafarAbdi (approved):

stephanie-eng:
Can we add a small reminder to this blurb in the pick_ik tutorial to install `pick_ik` before running the pre-configured example? We already have instructions in the tutorial on how to do it, but since it won't quite be "out-of-the box", wouldn't hurt to remind folks.
image

stephanie-eng (approved):

stephanie-eng:
If you don't want to do the above, I can probably just do it in a future PR later, I'm also fine with this as-is.

PR #1030, opened 2025-03-31
Fix two nodes launched with same name of /rivz2

Juliaj:
### Description

Address https://github.com/moveit/moveit2_tutorials/issues/991 where two nodes are launched with same name `rviz2`. The fix is to remove node name attribute when launching `rviz2`. 

### Testing 
- [x] Validated the updated Python files are launched successfully and there is no duplicate `rviz2` node.

### DIFF
--- a/doc/examples/jupyter_notebook_prototyping/jupyter_notebook_prototyping_tutorial.rst
+++ b/doc/examples/jupyter_notebook_prototyping/jupyter_notebook_prototyping_tutorial.rst
@@ -108,7 +108,7 @@ Once our MoveIt configuration is defined we start the following set of nodes:
                 package="rviz2",
                 executable="rviz2",
                 output="log",
-                arguments=["-d", rviz_config_file, "--ros-args", "--remap", "rviz:__name:=rviz2"],
+                arguments=["-d", rviz_config_file],
                 parameters=[
                         moveit_config.robot_description,
                         moveit_config.robot_description_semantic,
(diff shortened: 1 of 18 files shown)


sea-bass (commented):
Thanks! Syntax wise, this looks more complicated than just removing the `name` keyword arg, in that it does this but then reapplies it manually to one of the nodes using lower level args.

What is the difference in the node names between the original proposed approach and this? Unless there is a good reason, I'd opt for simpler syntax since these are tutorials for newcomers.

Juliaj:
> What is the difference in the node names between the original proposed approach and this? 

@sea-bass , thank you for the feedback. I agreed that the syntax is more complicated. Below are the differences in terms of `rviz2` node naming. My thoughts were to keep the node naming consistent with the package name. However, even with remapping rule, one of the node name still has `rviz` prefix. 

- Before (taken from original issue report), duplicated `rviz2` nodes.
```
/rviz2
/rviz2
/rviz2_private_133986940619904
/rviz2_private_133987075316320
```

- With name remapping change for `rviz2`. Three of the nodes renamed to `rviz2` which is aligned with package name `rviz2`.  The `rviz_visual_tools_gui` hasn't been renamed, some inconsistency. 

```
/rviz2
/rviz2_private_139289413291728
/rviz2_private_139289816608576
/rviz_visual_tools_gui
```

- With removing of `name="rviz2"`. `rviz2` related nodes have naming of `rviz` which is different than the package name. All `rviz2` nodes have the same naming prefix. 
```
/rviz
/rviz_private_127771787398864
/rviz_private_127772324975040
/rviz_visual_tools_gui
```

sea-bass:
Thanks @Juliaj, makes sense! I think the differences are small enough that I would favor the simpler code option, i.e., the last snippet in your comment. What do you think?

Juliaj:
> I think the differences are small enough that I would favor the simpler code option, i.e., the last snippet in your comment. What do you think?

Thanks @sea-bass , I'm good with that! Updated PR based on your suggestion.

rhaschke (approved):

sea-bass (approved):
Thanks for doing this!

PR #1036, opened 2025-04-11
Remove Google Analytics ID

sea-bass:
### Description

This was causing build errors on a recent PR, and also who cares about this?

Error:

```
Warning, treated as error:
The analytics_id option is deprecated, use the sphinxcontrib-googleanalytics extension instead.
```

I feel like instead of figuring this out or pinning back the Sphinx version, it's best to reMoveIt!

### Checklist

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -112,8 +112,6 @@ html_theme_options = {
     "collapse_navigation": True,
     "sticky_navigation": True,
     "navigation_depth": -1,
-    # This is the Google Analytics account used by moveit.ros.org, not picknik.ai
-    "analytics_id": "UA-108532843-1",
     # Only display the logo image, do not display the project name at the top of the sidebar
     "logo_only": True,
 }


mikeferguson:
Perhaps @EzraBrooks can find out who owns the analytics account and can offer thoughts on this?

sjahr (approved):

sea-bass:
> Perhaps @EzraBrooks can find out who owns the analytics account and can offer thoughts on this?

Good point. Also tagging @nbbrooks in case you guys would prefer to fix this vs. remove it altogether.

stephanie-eng (approved):
If we need to remove it for now to prevent errors, that's also OK by me. If there's desire to have it, we can restore when fixed.

sea-bass:
> If we need to remove it for now to prevent errors, that's also OK by me. If there's desire to have it, we can restore when fixed.

I'm okay waiting a few days for an answer since the code side of things still works.

PR #1039, opened 2025-04-16
Rolling: Add OMPL config to moveit_cpp.yaml

AndyZe:
It took me way too long to figure out why `moveit_cpp` was using the CHOMP planning pipeline. It's because `ompl` pipeline wasn't configured here, so MoveIt defaults to the first planning pipeline in the list.

(I've only tested this on Humble, hopefully it hasn't changed since then.)

### DIFF
--- a/doc/examples/moveit_cpp/config/moveit_cpp.yaml
+++ b/doc/examples/moveit_cpp/config/moveit_cpp.yaml
@@ -12,6 +12,19 @@ planning_pipelines:
   pipeline_names: ["ompl"]
 ompl:
   planning_plugin: ompl_interface/OMPLPlanner
+  request_adapters: >-
+    default_planner_request_adapters/AddTimeOptimalParameterization
+    default_planner_request_adapters/FixWorkspaceBounds
+    default_planner_request_adapters/FixStartStateBounds
+    default_planner_request_adapters/FixStartStateCollision
+    default_planner_request_adapters/FixStartStatePathConstraints
+  planner_configs:
+    PRMstarkConfigDefault:
+      type: geometric::PRMstar
+  # Define planner(s) for each move_group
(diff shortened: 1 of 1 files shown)


doc/examples/moveit_cpp/config/moveit_cpp.yaml
```diff
@@ -10,6 +10,8 @@ planning_scene_monitor_options:
 planning_pipelines:
   #namespace: "moveit_cpp"  # optional, default is ~
   pipeline_names: ["ompl"]
+ompl:
+  planning_plugin: ompl_interface/OMPLPlanner
```
AndyZe, line 15:
```suggestion
  planning_plugin: ompl_interface/OMPLPlanner
  # Add additional planning pipeline config here
```

sea-bass (changes requested):
Thanks, Andy! Unfortunately, it is different from Humble.

See for example [the latest `moveit_resources` Panda config](https://github.com/moveit/moveit_resources/blob/9e3c48401260a3d225903b901ecb596ce609219f/panda_moveit_config/config/ompl_planning.yaml#L1).

Specifically, notice the split of `request_adapters` and `response_adapters`, and that some of the adapters were renamed/removed.

Tagged @sjahr as his planning pipeline refactor from a little while back is what introduced this.

It may also be good to introduce something into this repo that actually tests the MoveItCpp pipeline modified here, because as far as CI is concerned, all is green.

AndyZe:
I'll open a new PR to target Humble, and update this one per the new planning pipelines. Prob won't have time for a new test though.

Sad to think how many thousands of people probably tried `moveit_cpp` and just failed silently.

sea-bass (approved):
Sweet, thanks!

PR #1045, opened 2025-05-21
Remove gripper_controllers dependency

fmauch:
The package got removed upstream in https://github.com/ros-controls/ros2_controllers/pull/1652. Find_packageing this on Kilted or Rolling will lead to errors from the next sync on.


sea-bass:
CI failures fixes in #1052 btw

PR #1063, opened 2025-08-22
Fix tf2*.h compile deprecations

mosfet80:
### Description

using tf2 *.h hesders is deprecated ->switch to *.hpp

### Checklist

## Summary by CodeRabbit

* **Chores**
  * Updated TF2 header references in the trajectory optimization example.
  * No user-facing behavior or runtime functionality changed.


MarqRazz (commented):
Sorry for the delay on getting you a review I was off traveling. I don't think we normally support distros directly in the code like this, can't we just not port this change to the Galactic and older branches?

mosfet80:
@MarqRazz I agree with your reasoning. I followed the method used in the moveit2 repository.

MarqRazz:
> @MarqRazz I agree with your reasoning. I followed the method used in the moveit2 repository.

If we are not going to port this back to Galactic why do we need the if else in the include?

MarqRazz (approved):

PR #1082, opened 2026-01-09
Correct naming and order of planning response adapters

niemsoen:
### Description

The AddRuckigTrajectorySmoothing was mentioned as a request adapter, even though it is a response adapter. The order of  response_adapters in the ompl_planning.yaml needs to be AddTimeOptimalParameterization first, then AddRuckigTrajectorySmoothing, which will correctly execute them in the order of (1) AddTimeOptimalParameterization, (2) AddRuckigTrajectorySmoothing.

I changed the order of the response adapters and correctly named them as response, not request adapters.

### Checklist


AndyZe (approved):

niemsoen:
@AndyZe Thank you for approving. Unfortunately the CI fails I believe for reasons that are not caused by this readme change. I see that the CI on main fails. As I'm not able to merge this I guess you will merge this at a later point when the CI works again?

AndyZe:
That's right. It's set to auto-merge now, as soon as the CI gets fixed. Thanks for the update.

